PR #7, opened 2018-10-29
Support Redux DevTools Extension when using SSR

m5r:


ctrlplusb:
Awesome catch! Thanks for this @m5r! 👍

PR #9, opened 2018-11-04
Add chapter about subscribe method to readme

Hurtak:


Hurtak:
Close in favour of #10

PR #10, opened 2018-11-04
Add chapter about subscribe to readme

Hurtak:


ctrlplusb:
Hey @Hurtak 👋

Thanks for raising this PR.

It's an interesting proposition as technically I am just abstracting Redux and do explicitly state so in the docs, including the fact that we return a standard Redux store - so their API docs should stand as a reference.

I do think it may be confusing to some though so I am going to consider this PR and see where we can best fit it in.

Your interest and help is appreciated. 🙏

Hurtak:
Hi,

good point that we just have redux-store so there is no need for Redux docs duplication. I read little more in the docs and now it makes more sense that this PR does not really fit that well in the docs.

PR #11, opened 2018-11-06
ci: test Node.js 6, 8, 10 and 11

DanielRuf:
Node.js 6, 8 and 10 are the current LTS branches and 11 is stable. See https://github.com/nodejs/Release/blob/master/README.md

We should test them all to catch any issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ node_js:
   - 'node'
   - '10'
   - '8'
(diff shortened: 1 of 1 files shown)


DanielRuf:
> We are essentially testing that the development experience of easy-peasy works across LTS versions of Node right

This + dependencies and the test framework.

PR #14, opened 2018-11-07
Disable devTools by default if NODE_ENV === production

christianchown:
and fix the typo in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -439,7 +439,7 @@ Creates a Redux store based on the given model. The model must be an object and 
 
     Provides custom configuration options for your store. It supports the following options:
 
-    - `devTool` (bool, not required, default=false)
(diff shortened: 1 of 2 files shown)


christianchown:
Ah - just read https://hackernoon.com/whats-new-with-server-side-rendering-in-react-16-9b0d78585d67#d14d which I didn't know about...

ctrlplusb:
Yeah, let's avoid any possible performance issues as it is very possible the process.env would be hit per request in an SSR environment. Fixing the property name definitely helps though!! 😊 

We could consider more docs around this to educate adopters and allow them to choose a performance strategy that suits them. 👍

christianchown:
Hi - the README now notes incorrectly that `devTools` default is true `if NODE_ENV !== 'production'` rather than `true`

ctrlplusb:
Woops. I'll fix. Ta

PR #18, opened 2018-11-08
Add expanded 'Usage with React Native' section to readme

### LINKED ISSUE #16: React Native has no native Symbol - polyfill or refactor?
Hello

React Native developers may run into this, as React Native has no native Symbol implementation: https://github.com/facebook/react-native/issues/4676

no-symbol

I had used the https://github.com/medikoo/es6-symbol polyfill in the app I used to test easy-peasy, and only discovered the problem when I took it out. 

If the effect/select etc. Symbols are in easy-peasy there will need to be a line about a polyfill being needed for Symbols in React Native in the README

christianchown:
Closes #16 

I've moved the _Usage with React Native_ to immediately follow the _Usage with React_ section, as it felt odd to have the API section sandwiched between the two.

Full readme here: https://github.com/christianchown/easy-peasy/tree/react-native-readme#readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -430,7 +430,7 @@ Easy Peasy is platform agnostic but makes use of language constructs and feature
 
 ### Polyfilling Symbol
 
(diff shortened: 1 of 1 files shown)


christianchown:
Also corrects location of `compose` config option

PR #31, opened 2018-11-17
Adds ability to enhance root reducer

danielking:
Adds ability to enhance root reducer for example when using redux-persist.


danielking:
:), Can't wait to have it. Please also consider about https://github.com/omodule/lazy-reducer.

PR #50, opened 2019-01-16
fix unused "endpoint" parameter in the generic helper example

skulptur:
This should fix the example by using the provided endpoint parameter


ctrlplusb:
Awesome, thanks @skulptur ♥️

Issue #53, opened 2019-01-23
need context export


takion:
Hello,

I need the ability to import context object to use with https://github.com/supasate/connected-react-router
wich rely by default on ReactReduxContext from react-redux but offer the possibility to provide a custom context and in this case, I need the one from easy-peasy.
Could you please just export it.
Many thanks (and also for making this great library).

ctrlplusb:
I recommend using an alternative lib that doesn't make you depend on react-redux.

Discussed here:
https://github.com/ctrlplusb/easy-peasy/issues/25#issuecomment-438053615

takion:
The problem is not react-redux, it's not using it, the problem is that this lib expect to extract store from a context consumer, so I have to provide the context object of easy-peasy, and that could be the case for others libraries too.
By the way, I don't know another library that connect react router to redux and I don't want to reinvent the wheel. Can you tell me what is your recommandation ?

ctrlplusb:
Well, they are depending on the context that is provided by the `react-redux` library. 😀So even if we exposed the context from `easy-peasy` it still would work with other libraries expecting the `react-redux` version (I think).

Good news though! There are alternatives available. 😀

As stated in the comment I linked to above you can use [`redux-first-history`](https://github.com/salvoravida/redux-first-history). I've used it myself with this library and it's great. 👍

takion:
I was just lookin on redux-first-history, thank you for the confirmation that is your recommandation with easy-peasy.

After checking the code of  connected-react-router, it seem that the only missing thing to work with is the context reference.

I don't understand why you mention connected-react-router in the documentation https://github.com/ctrlplusb/easy-peasy#reducerfn and why it's said:

> ... integrations with existing libraries, or legacy Redux code ...

And after there is an example that use connect decorator from react-redux, meaning for me that it's fully compatible. https://github.com/ctrlplusb/easy-peasy#finally-use-connect-against-your-components

takion:
Why not simply replace the easy-peasy context by redux 6 context using
`import { ReactReduxContext as StoreContext } from 'react-redux'`

ctrlplusb:
Reopening for more discussion and clarity. Will respond soon. 👍

strobox:
Maybe it is not a problem to provide two context, one from *react-redux*:
```es6
import { Provider } from 'react-redux'
import { StoreProvider } from 'easy-peasy';
export default hot(() => <StoreProvider store={store} ><Provider store={store} >
[...]

PR #54, opened 2019-01-23
react-redux interop + redux-persist compat + Symbol

takion:
see issue https://github.com/ctrlplusb/easy-peasy/issues/53

### DIFF
--- a/src/easy-peasy.js
+++ b/src/easy-peasy.js
@@ -4,15 +4,25 @@ import {
   createStore as reduxCreateStore,
 } from 'redux'
 import memoizeOne from 'memoize-one'
-import produce from 'immer'
+import produce, { setAutoFreeze } from 'immer'
 import thunk from 'redux-thunk'
 import { isStateObject } from './lib'
 
-const effectSymbol = '__effect__'
-const selectSymbol = '__select__'
-const selectDependenciesSymbol = '__selectDependencies__'
-const selectStateSymbol = '__selectState__'
-const reducerSymbol = '__reducer__'
+setAutoFreeze(false)
+
+const effectSymbol = Symbol('__effect__')
+const selectSymbol = Symbol('__select__')
+const selectDependenciesSymbol = Symbol('__selectDependencies__')
+const selectStateSymbol = Symbol('__selectState__')
+
+class reducerContainer {
+  constructor(fn) {
+    this.fn = fn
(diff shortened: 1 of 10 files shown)


src/easy-peasy.js
```diff
@@ -4,15 +4,25 @@ import {
...
+import produce, { setAutoFreeze } from 'immer'
 import thunk from 'redux-thunk'
 import { isStateObject } from './lib'
 
-const effectSymbol = '__effect__'
-const selectSymbol = '__select__'
-const selectDependenciesSymbol = '__selectDependencies__'
-const selectStateSymbol = '__selectState__'
-const reducerSymbol = '__reducer__'
+setAutoFreeze(false)
```
ctrlplusb: Interesting. I had no idea about this. Perhaps we should tie the value to the dev boolean config of the store?

src/easy-peasy.js
```diff
@@ -4,15 +4,25 @@ import {
...
+
+const effectSymbol = Symbol('__effect__')
```
ctrlplusb: We had issues with Symbols - see #16

src/easy-peasy.js
```diff
@@ -4,15 +4,25 @@ import {
...
+const selectSymbol = Symbol('__select__')
+const selectDependenciesSymbol = Symbol('__selectDependencies__')
+const selectStateSymbol = Symbol('__selectState__')
+
+class reducerContainer {
```
ctrlplusb: Whats the thinking behind having this container?

src/easy-peasy.js
takion: I agree, that would be better

src/easy-peasy.js
takion: Ah ok, maybe you could take the same approach that I took with reducerContainer, a wrapper and then instanceof, to let clean the original references. I think adding property, and especially this way, to parameters will polute them.

src/easy-peasy.js
takion: Compatibility with redux-persist that use Object.freeze api and then lead to an error whan easy-peasy try to modify it, even with adding a Symbol

ctrlplusb:
Closing this as won't merge. But did take your hint at the object freeze disable. Thanks! ❤️

PR #33, opened 2018-11-21
Add Typescript typings

christianchown:
Closes #21 

Adds the following types to define a Model for the store
- Effect<Model, _optional_ Payload, _optional_ ReturnType>
- Action<StoreValues, _optional_ Payload>
- Reducer\<State\>
- Select<StoreValues, ResultantType>

Adds the following types to extract useful shapes from a Model
- ModelValues\<Model\>
- MutableModelValues\<Model\>
   _store values that can be mutated by actions i.e. ones not from `select(...)`)_ or `reducer(...)`
- ModelActions\<Model\>

Adds the following types for use in `createStore`
- Config\<Model\>
- Store\<Model\>

Adds the following type for use in `useAction`
- Dispatch\<Model\>

And adds type information for the public API
- createStore\<Model\>
- effect\<Model, _optional_ Payload, _optional_ ReturnType\>
- reducer\<State\>
- select\<StoreValues, ResultantType\>
- StoreProvider\<Model\>
- useStore(mapState: ModelValues\<Model\> => StateValue)
- useAction(dispatch: Dispatch\<Model\> => ActionPayload)

The types in the public API are opt-in, and will default to `any` if no generics are provided.

```typescript
import React from 'react';
import {
  createStore,
  effect,
  reducer,
  select,
  useStore,
  useAction,
  Action,
  Effect,
  Reducer,
  Select,
  ModelValues,
  Dispatch,
} from 'easy-peasy';
import { Action as ReduxAction } from 'redux';
import todoService from './toDoService';

interface TodoValues {
  items: Array<string>;
}

interface TodoValuesAndSelectors extends TodoValues {
  lengthOfItems: number;
}

interface TodoActions {
  saveTodo: Effect<Model, string, Promise<number>>;
  todoSaved: Action<TodoValuesAndSelectors, string>;
  lengthOfItems: Select<TodoValuesAndSelectors, number>;
}

interface Model {
  todos: TodoValues & TodoActions;
  counter: Reducer<number>;
}

const store = createStore<Model>({
  todos: {
    items: [],

    saveTodo: effect(async (dispatch, payload, getState) => {
      const saved = await todoService.save(payload);
      dispatch.todos.todoSaved(saved); // 👍 correctly typed

      // dispatch.todos.todoSaved(1);
      // 👍 errors! (1 is not assignable to string)

      // dispatch.notToDos.something();
      // 👍 errors! (notToDos does not exist on Dispatch<Model>)

[...]

PR #57, opened 2019-01-25
TypeScript definitions

ctrlplusb:
__NOTE__: We have been iterating on the types for the lifetime of this PR. Changes are being made, but I am updating this comment to reflect the latest version of the definitions. 

Your help in testing them would be greatly valued! 👍

---

Firstly, I must absolutely stress that none of this work would have been achieved without the monumental effort of @christianchown on his typings work via #33! To be honest given the API of `easy-peasy` I never considered that reasonably sound typings would be at all possible. @christianchown proved me wrong with his work. 

Originally I tried to see if I could solve the issue that was being faced in #33 regarding being bound to a maximum of 4 levels deep of typed state. In doing so however I began to spin out a solution which had a significantly different type definition implementation to that of #33. Eventually it diverged so much that I eventually settled on the idea of spinning out a completely alternative approach rather than trying to change #33 too much.

So I have created this PR instead of trying to comment over #33. I hope that we can take the best of the two to come up with a final solution we are all happy with.

The core approach used in these typing is the use of generic recursive type aliases. I discovered the concept by chance after many hours of googling. The approach appears to work very well for our case, however, I am not sure on how it will perform against large model definitions.

This PR also differs in the manner by which you define your model. Given that I was able to leverage generic recursive type aliases I was able to allow state, actions, selectors, and reducers to be collocated.

Below is a pretty comprehensive example.

```typescript
import * as React from 'react'
import * as ReactDOM from 'react-dom'
import {
  createStore,
  effect,
  reducer,
  select,
  StoreProvider,
  useAction,
  useStore,
  Action,
  Dispatch,
  Effect,
  Reducer,
  Select,
  State,
} from 'easy-peasy'
import { connect } from 'react-redux'

/**
 * Firstly you define your Model
 */

interface TodosModel {
  items: Array<string>
  firstItem: Select<TodosModel, string | void>
  addTodo: Action<TodosModel, string>
}

interface UserModel {
  token?: string
  loggedIn: Action<UserModel, string>
  login: Effect<Model, { username: string; password: string }>
}

interface Model {
  todos: TodosModel
  user: UserModel
  counter: Reducer<number>
}

/**
 * Then you create your store.
 * Note that as we pass the Model into the `createStore` function, so all of our
 * model definition is typed correctly, including inside the actions/helpers etc.
 */

const store = createStore<Model>({
  todos: {
    items: [],
    firstItem: select(state =>
      state.items.length > 0 ? state.items[0] : undefined,
    ),
    addTodo: (state, payload) => {
      state.items.push(payload)
    },
  },
  user: {
    token: undefined,
    loggedIn: (state, payload) => {
      state.token = payload
    },
    login: effect(async (dispatch, payload) => {
      const response = await fetch('/login', {
        method: 'POST',
        body: JSON.stringify(payload),
        headers: {
          'Content-Type': 'application/json',
        },
      })
      const { token } = await response.json()
      dispatch.user.loggedIn(token)
    }),
  },
  counter: reducer((state = 0, action) => {
    switch (action.type) {
      case 'COUNTER_INCREMENT':
        return state + 1
      default:
        return state
    }
  }),
})

/**
 * You can use the "standard" store APIs
 */

console.log(store.getState().todos.firstItem)

store.dispatch({ type: 'COUNTER_INCREMENT' })

store.dispatch.todos.addTodo('Install typescript')

/**
 * You can access state via hooks
 *
 * FYI - see @formula349's tip so that you don't have to constantly provide
 *       types on your useStore and useAction hooks:
 *       https://github.com/ctrlplusb/easy-peasy/issues/21#issuecomment-457644655
 */
function MyComponent() {
  const token = useStore((state: State<Model>) => 
    state.user.token
  )

  const login = useAction((dispatch: Dispatch<Model>) => 
	dispatch.user.login,
  )

  return (
    <button onClick={() => login({ username: 'foo', password: 'bar' })}>
      {token || 'Log in'}
    </button>
  )
}

/**
 * Expose the store to your app as normal
 */

ReactDOM.render(
  <StoreProvider store={store}>
    <MyComponent />
  </StoreProvider>,
  document.getElementById('root'),
)

/**
 * We also support typing react-redux
 */

const Counter: React.SFC<{ counter: number }> = ({ counter }) => (
  <div>{counter}</div>
)

connect((state: State<Model>) => ({
  counter: state.counter,
}))(Counter)
```

For anyone interested in testing these you can install:

```
npm install easy-peasy@typescript
```

Things I dislike or need to be considered about these typings:

 - [ ] Having to explicitly add the typing for any injections to any Effect (`Effect<Model, void, Injections>`). It would be great to have an intermediary type that allows us to bake in the value for all effects (`type EffectWithInjections = Effect<*, *, Injections>`) and then be able to declare effects with the new type. 
 - [ ] Unknown performance properties. The recursive type resolution may bite us. We should try testing against a significantly large definition to ensure this isn't a blocker. 

@christianchown - please feel free to merge any bits of this into your PR. I am not preferring one over the other for now, but it may be useful to evolve two forms until we can eventually merge into a single solution that is effective.

Like I said above I still have concerns about the perf implications of the design of these types.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,7 +1,8 @@
 {
   "presets": [
     "@babel/preset-env",
-    "@babel/preset-react"
+    "@babel/preset-react",
+    "@babel/preset-typescript"
   ],
   "plugins": [
     "@babel/plugin-syntax-dynamic-import",
--- a/index.d.ts
+++ b/index.d.ts
@@ -50,79 +50,73 @@ type EffectMeta = {
  * This type recursively filters a model down to the properties which
  * represent actions
  */
-export type Actions<TModel extends Object> = {
+export type Actions<Model extends Object> = {
   [P in keyof (Omit<
-    TModel,
-    KeysOfType<
-      Pick<TModel, KeysOfType<TModel, Object>>,
-      NaturalState | UtilTypes
-    >
-  >)]: Actions<TModel[P]>
+    Model,
+    KeysOfType<Pick<Model, KeysOfType<Model, Object>>, NaturalState | UtilTypes>
+  >)]: Actions<Model[P]>
 } &
   {
-    [P in keyof Pick<TModel, KeysOfType<TModel, ActionTypes>>]: Param1<
-      TModel[P]
+    [P in keyof Pick<Model, KeysOfType<Model, ActionTypes>>]: Param1<
+      Model[P]
     > extends void
-      ? () => UnsafeReturnType<Param1<TModel[P]>>
-      : (payload: Param1<TModel[P]>) => UnsafeReturnType<Param1<TModel[P]>>
+      ? () => UnsafeReturnType<Param1<Model[P]>>
+      : (payload: Param1<Model[P]>) => UnsafeReturnType<Param1<Model[P]>>
   }
 
 /**
  * This type recursively filters a model down to the properties which
  * represent state - i.e. no actions/selectors etc.
  */
-export type State<TModel extends Object> = {
+export type State<Model extends Object> = {
   [P in keyof (Omit<
-    TModel,
-    KeysOfType<
-      Pick<TModel, KeysOfType<TModel, Object>>,
-      NaturalState | UtilTypes
-    >
-  >)]: State<TModel[P]>
+    Model,
+    KeysOfType<Pick<Model, KeysOfType<Model, Object>>, NaturalState | UtilTypes>
+  >)]: State<Model[P]>
 } &
-  { [P in keyof Pick<TModel, KeysOfType<TModel, NaturalState>>]: TModel[P] } &
+  { [P in keyof Pick<Model, KeysOfType<Model, NaturalState>>]: Model[P] } &
   {
     readonly [P in keyof Pick<
-      TModel,
-      KeysOfType<TModel, Select<any, any>>
-    >]: UnsafeReturnType<TModel[P]>
+      Model,
+      KeysOfType<Model, Select<any, any>>
+    >]: UnsafeReturnType<Model[P]>
   } &
   {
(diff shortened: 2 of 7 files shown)


ctrlplusb:
Hmm, seeing some issues here...

ctrlplusb:
FYI, any updates to this PR are being immediately published.

```
npm install easy-peasy@typescript
```

formula349:
It doesn't seem the Payload generic parameter is used in the useAction function. Can this be removed?
```typescript
    export function useAction<
        Model extends Object = {},
        Payload = any,
        Result = any
    >(mapAction: (actions: Dispatch<Model>) => Result): Result
```

aalpgiray:
I am not sure if I have right configuration but here I have type any for state.items

``` typescript
 todos: {
    items: [],
    firstItem: select(state =>
      state.items.length > 0 ? state.items[0] : undefined
    )
  }
```

image

ctrlplusb:
@aalpgiray - good spot those deps need to be moved from "devDependencies" into "dependencies" 👍

@christianchown - nice, agreed. 🙏

ctrlplusb:
New version published

christianchown:
Hi @aalpgiray - here's the effect return bit from the docs https://github.com/ctrlplusb/easy-peasy/blob/master/README.md#dispatching-an-effect-action-directly-via-the-store

aalpgiray:
Makes sense in that regard. But I believe an effect should not return anything at all. But, it's a little too opinionated maybe. Your function changing are state of you model but olso returns something else, it is just bad pattern. Of course, this is typescript related topic and this talk is not however.

PR #59, opened 2019-01-29
Changes effect dispatching to have a "started" and "completed" action

### LINKED ISSUE #52: How to execute code once an action has updated the store?
I am trying to execute code once the store has been updated from a model action, and I do not see an easy way to subscribe to this. 

If I trigger the action in an effect, I am still seeing the effect promise resolving before the triggered action has updated the store. Is there any kind of event we can hook directly into to the action callback or store to achieve this?

Here is some example code of what I am facing:

```
const defaultState = {
  user: {},
  counter: {},
  ....
}

// model
const model = {
  ...defaultState,
  // Action
  resetState: (state) => ({ ...state, ...defaultState }),
  // Effect
  logOut: effect(async dispatch => {
    const asyncCall = await ApiService.logOut();
    return dispatch.resetState();
  })
}

// View
const logOutButton = () => {
  const logOut = useAction(dispatch => dispatch.logOut);

  return <Button 
                onClick={() => logOut().then(() => console.log('Assume that the store has updated from the resetState action'))}
               >Log Out</Button>
}
```

In the above psuedo code, I am seeing the console.log function in the logout button firing before the store has finished updating from the `resetState` action. In the devtools I am actually seeing the resolved promise function being invoked before the `resetState` action has been triggered.

Any help would be appreciated.

ctrlplusb:
Closes #52 

Also ensures that the actions indicating an effect always run. Previously it only executed in dev mode.

FYI, this is a breaking change, but only if you had custom listeners against effects. But then again you would have had to be deploying `easy-peasy` in dev mode for that to work in production. :)

### DIFF
--- a/src/__tests__/easy-peasy.test.js
+++ b/src/__tests__/easy-peasy.test.js
@@ -599,22 +599,32 @@ describe('internals', () => {
 })
 
 describe('effects', () => {
-  test('dispatches an action to represent the start of an effect', async () => {
+  test('dispatches an action to represent the start and end of an effect', async () => {
     // arrange
     const model = {
       foo: {
-        doSomething: effect(() => undefined),
+        counter: 0,
+        increment: state => {
+          state.counter += 1
+        },
+        doSomething: effect(dispatch => {
+          dispatch.foo.increment()
+          return 'did something'
+        }),
       },
     }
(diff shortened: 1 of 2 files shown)


ctrlplusb:
The effect action is to be deprecated.

PR #82, opened 2019-02-11
Fix typo in typings

bassplayerch:
Middlewares should be middleware.


ctrlplusb:
Nice spot @bassplayerch ❤️

PR #88, opened 2019-02-12
Adds test helpers and docs

ctrlplusb:
Closes #26 #69

PR #89, opened 2019-02-12
Remove setTimeout and fixup tests

ifyoumakeit:
Thanks for all your hard work on this, loving easy-peasy a lot.

In Chrome I'm seeing a bunch of setTimeout timing warnings due to setTimeout wrapping state setting. I saw your comment above it, but is there a test showing what happens when it is not wrapped? I started to make an issue about this, but thought a PR would be more helpful. 

I removed it and ran the tests and it failed on `maps state when prop dependency changes`.  To fix it I wrapped the second re-render in an `act` call to ensure that the effects are flushed before checking the DOM, and only checked it for the new value. I'm pretty sure this is the correct way to test with hooks, but I could be wrong.

## Other notes
- React is also throwing an message about the Counter components needing `act`.
- I think all of the `jest.runAllTimers()` can also be removed, but didn't want to create a large changeset without first understanding their need.

## Chrome warnings
screen shot 2019-02-12 at 5 35 17 pm

## React message during testing
```
 console.error node_modules/react-dom/cjs/react-dom.development.js:506
 Warning: An update to Counter inside a test was not wrapped in act(...).
      
When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
      
This ensures that you're testing the behavior the user would see in the browser. Learn more at https://fb.me/react-wrap-tests-with-act
````

### DIFF
--- a/src/__tests__/easy-peasy.test.js
+++ b/src/__tests__/easy-peasy.test.js
@@ -94,16 +94,14 @@ describe('react', () => {
     expect(value.firstChild.textContent).toBe('foo')
 
     // act
-    const app2 = (
-      <StoreProvider store={store}>
-        <Values id={2} />
-      </StoreProvider>
-    )
-
-    rerender(app2)
-
-    // assert
-    expect(value.firstChild.textContent).toBe('foo')
+    act(() => {
+      const app2 = (
+        <StoreProvider store={store}>
+          <Values id={2} />
+        </StoreProvider>
+      )
+      rerender(app2)
+    })
 
     // ensure settimeouts fire
     act(() => {
--- a/src/hooks.js
+++ b/src/hooks.js
@@ -25,16 +25,10 @@ export function useStore(mapState, dependencies = []) {
         return
       }
       stateRef.current = newState
-      // The settimeout wrap fixes a strange issue where a setState would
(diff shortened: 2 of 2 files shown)


ctrlplusb:
Hey @ifyoumakeit 

Thanks for your kind words - I'm glad that you are enjoying the library. Even more thanks for your contributions - very helpful! ❤️

Reviewing this now I am struggling to think of the exact case that caused me to do the `setTimeout` wrap. I am wondering if it had something to do with my early misunderstanding of hooks and the nature of closures.

I am going to review this ASAP and get back to you. I want to do an experiment with it prior to merging.

Thanks again 👍

ctrlplusb:
Wow, actually hit friction with this in doing my testing helpers. The setTimeout definition needs to be removed for testing experience to work as one would expect it to.  I have published an alpha on tag `next` with this applied. You were correct in not needing the run timers too 👍

ctrlplusb:
Yep, tested, and it's far better. Thanks for raising this.

I am going to close this in favour of #88 which incorporates your changes.

🙏

ifyoumakeit:
Great to hear and glad I could help! The testing docs are great.

PR #90, opened 2019-02-13
Fix issue that causes a thunk to remain uncompleted when err thrown

cevr:
I was making a middleware to dispatch loading states, and I found that thunks that threw an error were not marked as complete. This caused some of my state to remain unchanged when making an HTTP request that resulted in an error.

### DIFF
--- a/src/easy-peasy.js
+++ b/src/easy-peasy.js
@@ -225,7 +225,7 @@ export const createStore = (model, options = {}) => {
               })
               .catch(err => {
                 references.dispatch({
-                  type: `${actionName}(completed)`,
+                  type: `${actionName}(failed)`,
                   payload: err,
                 })
               })
--- a/src/easy-peasy.js
+++ b/src/easy-peasy.js
@@ -223,6 +223,12 @@ export const createStore = (model, options = {}) => {
                 })
                 return result
               })
+              .catch(err => {
+                references.dispatch({
+                  type: `${actionName}(completed)`,
+                  payload: err,
+                })
+              })
           actionCreator[actionNameSymbol] = actionName
           actionCreatorDict[actionName] = actionCreator
           set(path, actionCreators, actionCreator)
--- a/src/easy-peasy.js
+++ b/src/easy-peasy.js
@@ -238,6 +238,7 @@ export const createStore = (model, options = {}) => {
                   type: `${name}(failed)`,
                   payload: err,
                 })
+                return Promise.reject(err)
               })
 
           actionCreator[actionNameSymbol] = name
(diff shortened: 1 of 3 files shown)


ctrlplusb:
Hey @cevr 

Thanks for contributing!

I'm not sure if I like the semantics around marking a thunk as complete, even though it didn't complete. Currently the `listen` API specifies that it will only fire when the target action completes _successfully_, so this change would break that behaviour.

In my opinion there are two ways to handle this case:

1. Developers should ensure that their thunks have error handling built in
2. We could catch and dispatch a different action type when a thunk fails (e.g. `@thunk.foo.doSomething(failed)`). This would be helpfully explicit when trying to debug failures and perhaps the payload could even contain some error information. We could also use this to maintain the correct behaviour around the `listen` helper.

Thoughts?

cevr:
> Hey @cevr
> 
> Thanks for contributing!
> 
> I'm not sure if I like the semantics around marking a thunk as complete, even though it didn't complete. Currently the `listen` API specifies that it will only fire when the target action completes _successfully_, so this change would break that behaviour.
> 
> In my opinion there are two ways to handle this case:
> 
> 1. Developers should ensure that their thunks have error handling built in
> 2. We could catch and dispatch a different action type when a thunk fails (e.g. `@thunk.foo.doSomething(failed)`). This would be helpfully explicit when trying to debug failures and perhaps the payload could even contain some error information. We could also use this to maintain the correct behaviour around the `listen` helper.
> 
> Thoughts?

Yeah I see what you mean. I think we should follow the JavaScript semantics and dispatch a failed state.

Thanks @ctrlplusb  😃

src/easy-peasy.js
```diff
@@ -233,6 +233,13 @@ export const createStore = (model, options = {}) => {
                 )
                 return result
               })
+              .catch(err => {
+                references.dispatch({
+                  type: `${name}(failed)`,
+                  payload: err,
+                })
```
ctrlplusb: Maybe this should throw to propagate the error?

```javascript
              .catch(err => {
                references.dispatch({
                  type: `${name}(failed)`,
                  payload: err,
                })
                throw err;
              })
```

Which would result in a test result of:

```javascript
    expect(trackActions.actions).toEqual([
      { type: '@thunk.foo.doSomething(started)', payload },
      { type: '@thunk.foo.error(started)', payload: undefined },
      { type: '@thunk.foo.error(failed)', payload: undefined },
      { type: '@thunk.foo.doSomething(failed)', payload },
    ])
```

cevr: True! Good catch. Thanks @ctrlplusb

ctrlplusb:
Thanks for your work on this @cevr - I am going to base this against the V2 branch as technically it is a breaking change. 👍

PR #91, opened 2019-02-13
change deprecated effect to thunk

homura:


ctrlplusb:
Good spot, and thanks for the help ❤️

PR #98, opened 2019-02-15
[WIP] V2 - Adds ability to dynamically modify store model

### LINKED ISSUE #29: Lazy reducers

ctrlplusb:
Closes #29


ctrlplusb:
Got it working in principal, but have no idea how to represent this via the type system.

Needs a massive code cleanup still too. 😊

cevr:
I see that the API only supports adding one model at a time. Is there something limiting adding multiple models in one object as is done in the root model?

ctrlplusb:
Hey @cevr - no restriction per se, but felt that models would be added in such a manner - i.e. a model namespace to represent a specific branch of your application.

Do you have experience with this type of thing? If so I would love your feedback on why you think being able to merge multiples keys in a single action would be valuable. 👍

psachs21:
@ctrlplusb I've dealt with the typing issue before, unfortunately, the only thing we do is accept the eventual FULL state interface as a type param even if the initial model doesn't satisfy. There is no way to statically know if a model has been loaded or not. This leads to potentially accessing models that don't exist yet, but I think that is just a limit of static typing.

cevr:
> Hey @cevr - no restriction per se, but felt that models would be added in such a manner - i.e. a model namespace to represent a specific branch of your application.
> 
> Do you have experience with this type of thing? If so I would love your feedback on why you think being able to merge multiples keys in a single action would be valuable. 👍

Sorry, not that I have any experience myself. I was asking out of curiosity to learn more! 🙇

PR #130, opened 2019-03-21
Update README according to easy-peasy v2 API

ivosh:
Listeners now need to be declared with either `action` or `thunk`.


ctrlplusb:
Thank you! ❤️

PR #137, opened 2019-03-25
Lifts redux store subscriptions to context

### LINKED ISSUE #136: react: state mapper and render are called with stale props
Hi,

not sure if you are following the discussions about the [official react-redux hooks](https://github.com/reduxjs/react-redux/issues/1179), but one major issue we are trying to solve there is what has been called the "zombie child problem". Basically, it means that the selector in `useStore` and worst case even the render of the component happens with an inconsistent store and props. Sadly, your library suffers from the same issue with both cases happening, i.e. `useStore` and the render are executed with inconsistent state and props. You can see an example that causes a crash [here](https://codesandbox.io/s/7yyx72z8r6) and another example that logs what is happening to the console more clearly [here](https://codesandbox.io/s/oxoq8q06v5).

ctrlplusb:
Closes #136

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "easy-peasy",
-  "version": "2.1.1",
+  "version": "2.2.0-alpha.2",
   "description": "Easy peasy global state for React",
   "license": "MIT",
   "main": "dist/easy-peasy.cjs.js",
--- a/src/__tests__/issues.test.js
+++ b/src/__tests__/issues.test.js
@@ -1,0 +1,54 @@
+import React from 'react';
+import { act } from 'react-dom/test-utils';
+import { render } from 'react-testing-library';
+
+import { action, createStore, useStore, StoreProvider } from '../index';
+
+test('issue#136', async () => {
+  // arrange
+  const store = createStore({
+    items: {
+      a: { id: 'a', name: 'A' },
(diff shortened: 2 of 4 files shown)


MrWolfZ:
Regarding your comment [here](https://github.com/reduxjs/react-redux/issues/1179#issuecomment-476267274):

> Hi all, joining this discussion as @MrWolfZ informed me that my library would suffer the same "zombie child" issue as discussed here.
>
> I've managed to get around this by lifting the redux store subscriptions for the hooks to the root context, which allows me to execute the mapStates in an order that is more likely to represent the component hierarchy.
>
> PR is here: ctrlplusb/easy-peasy#137
>
> Demo: https://codesandbox.io/s/pm39mzn13m
>
> I still need to do a lot more tinkering. But wanted to put this on your radar. Hoping to get some early feedback on whether this is a crazy idea or not. 👍

Lifting the subscriptions to the root context does nothing to the order in which `mapState`s are executed. Also, you are using `Object.values` which is even worse, since it has an undefined ordering. Also, as mentioned in #136 this code throws hidden errors.

ctrlplusb:
Appreciate the review ❤️

ctrlplusb:
Closing as better understand the problem now and have an alternative solution brewing...

PR #160, opened 2019-04-30
Note about nested state in docs

bentron2000:
Just a suggestion to clarify how nested state works in easy-peasy - Thanks! B


ctrlplusb:
Thanks for this @bentron2000 - I am doing a bunch of work on the docs in the background and will keep this in mind 👍

Issue #170, opened 2019-05-13
Example for derived state with dependencies and arguments


johnnythunder:
Could you please provide an example of composing selectors with arguments?

```javascript
const store = createStore({
  products: [{ id: 1, name: 'Shoes', price: 123 }, { id: 2, name: 'Hat', price: 75 }],

  productById: select(state =>
    id => state.products.find(x => x.id === id)
  ),

  // should be updated to use productById as a dependency
  productPrice: select(state =>
    id => '$' + state.products.find(x => x.id === id).price
  )
};
```

richvida:
This would be useful to know for me as well.

ctrlplusb:
Eeek, some bugs around selectors when I tried to implement the solution. Will get back to you with an example when working.

ctrlplusb:
FYI both - I have come up with a much better architectural design for selectors that will address your issues, and be far more performant. Plus you won't have to do this nested function based implementation which just sucks.

I am hoping to have time this weekend to be able to implement it. 👍

ctrlplusb:
FYI I have published a new helper. Check out #177 for more info. I've still got loads of docs to write out. I'll be writing how to use them via Typescript too after I finish the standard JS examples.

Would really appreciate your feedback.

ctrlplusb:
This is now in the [next](#181) branch

ctrlplusb:
FYI, there was a separate bug on selectors which has been fixed. You can call selects from each other now. 👍

https://codesandbox.io/s/easypeasy170-l7dle

I am still of mind to deprecate `select` in favour of `selector` - despite it being more verbose. It's still less of a footgun.

PR #177, opened 2019-05-17
Adds new derived helper

### LINKED ISSUE #175: How select access select or state in other subtrees without put the select root
the use-case

```js
const store = createStore({
  authors: {
    data: {
      1: { name: "John Smith" }
    }
  },
  posts: {
    data: {
      1: { authorId: "1", title: "post title 1" },
      2: { authorId: "1", title: "post title 2" }
    },
    detail: select(posts => postId => ({
      ...posts.data[postId],
      author: ?? // how to reference the author in subtree of store
    }))
  }
});

```

ctrlplusb:
This PR has been published as v2.3.0-alpha.0, under the `next` tag. You can install it like so:

```
npm install easy-peasy@next
```

Closes #175 #170 

I have some basic docs below, but they still need to be fleshed out a bit more. I will be writing docs for Typescript usage too. All of the below supports full typing too. You can look at the typescript definition tests in this branch if you are keen to see some examples.

### Intro

This PR takes all of my learnings from the current selector API and introduces a new derived state alternative API, named `derived`, that attempts to resolve the problems and limitations that I identified with the `select` helper.

When this gets merged, the `select` API will be marked as deprecated. Any use of it will output a warning message, advising that existing implementations migrate to the new helper. 

When the next major release of `easy-peasy` happens we will then remove the deprecated `select` helper.

### Documentation

The new `derived` helper is intended to allow you to derive state based on your existing state. It performs memoization under the hood and only attempts to resolve the derived state again when the inputs to the selector change.

**Simple Example**

Firstly, we will demonstrate a very straight forward implementation.

```javascript
import { derived } from 'easy-peasy';

const store = createStore({
  todos: {
    items: [],
    count: derived(
      // The first argument to derived requires you to provide an array of
      // "argument selectors". These "argument selectors" are functions that
      // will take in the state of the model they are bound against, and will
      // return the part of the state required by your derived state selector.
      // The results of each of the "argument selectors" become the arguments
      // to your derived state selector.
      [(state) => state.items], 
      
      // Now you defined your selector. It receives the arguments, which are 
      // a result of your argument selectors above.
      // You then return the derived state based on the arguments
      (items) => items.length,
    }),
  }
});
```

The above is slightly more verbose than the `select` API, however, by forcing you to first select the parts of your model your derived state selector cares about (via the "argument selectors") we are able to provide much better performance characteristics for you.

Consumption of your derived state has also changed. You no longer get the derived state directly available via the `useStore` hook. Instead a function is exposed at the same path allowing you to request the derived state.

For example:

```javascript
import { useStore } from 'easy-peasy';

function TotalTodos() {
  const todoCount = useStore(state => state.todos.count);
  // Notice how we execute the todoCount below:
  //                       👇
  return <div>Total: {todoCount()}</div>;
}
```

There are a few reasons we have moved to this API design:

1. It provides us with much improved performance characteristics, as derived state selectors will only be resolved/executed on demand - and will subsequently be cached, with the cached result available to any subsequent use of the derived state within your app.
2. It opens up the opportunity for us to support "runtime arguments" to your derived state selectors - I will explain this further below.

Also, be aware, that your component will rerender if changes are made to the underlying state that your derived state selector depends on. So in our example above our component would rerender if a todo item was added or removed. Our component will only rerender in this condition - no other state updates to the todos model or other parts of your store will cause any rerender.

**Accessing global state from your "argument selectors"**

A fairly common request we have had was the ability to select across different branches of your model. With the `select` helper this was very hard to do, especially in a performant manner. The new `derived` helper provides a simple API, with optimisation characteristics that match using local state.

```javascript
const store = createStore({
  todos: {
    items: {
      1: { id: 1, text: 'Win the lottery' ]
    }
  },
  profile: {
    favouriteTodoId: 1,
    favouriteTodo: derived(
     // Notice how we are providing multiple "argument selectors" now
     [
       (state) => state.favouriteTodoId,
       // Our argument selectors receive as a second argument the entire
       // store state
       (_, storeState) => storeState.todos.items
     ],
     // We receive both our argument selectors and do the derived state
     (todoId, todos) => todos[todoId]
    )
  }
});
```

**Example of runtime arguments within your derived state**

Your derived state helpers support runtime arguments. Any arguments that are provided to your derived state selector during access will be passed into it. This is best illustrated with an example.

```javascript
const store = createStore({
  todos: {
    items: {
      1: { id: 1, text: 'Win the lottery' ]
    },
    getById: derived(
      [(state) => state.items], 
      // Note this second argument. It wasn't defined above. We therefore
      // expect it to be provided at runtime
      //      👇
      (items, id) => items[id]
  },
});
```

And here is an example using it:

```javascript
import { useStore } from 'easy-peasy';

function Todo({ id }) {
  const getTodoById = useStore(state => state.todos.getById);
  // Notice how we provide the runtime argument
  //                       👇
  const todo = getTodoById(id);
  return todo ? <div>{todo.text}</div> : null;
}
```

**Customising the cache limit for derived state (especially helpful when runtime args used)** 

By default the derived state has a cache limit of 1. This means that if you provide a new runtime argument every time you hit your derived state helper it will be run.

Typically when you are introducing a derived state selector that consumes runtime arguments you will want to customise the cache limit. Imagine rendering a list of todos, where you called the `getTodoById` multiple times with each id. You certainly want to have at least a deeper limit of caching to support this and avoid unnecessary re-renders.

You can easily customise the cache limit via the 3rd argument to the `derived` helper. The config object.

```javascript
const store = createStore({
  todos: {
    items: {
      1: { id: 1, text: 'Win the lottery' ]
    },
    getById: derived(
      [(state) => state.items], 
      (items, id) => items[id],
      // We will cache the most recent 100 unique calls to 
      // our selector based on the runtime arguments
      //        👇
      { limit: 100 }
  },
});
```

**Derived state selectors referencing each other**

You can reference selectors from any part of your tree when defining one. It follows much of the same principals as consuming other state.

```javascript
const store = createStore({
  todos: {
    items: {
      1: { id: 1, text: 'Win the lottery' ]
    },
    getById: derived(
      [(state) => state.items], 
      (items, id) => items[id],
      { limit: 100 }
  },
  },
  profile: {
    favouriteTodoId: 1,
    favouriteTodo: derived(
     // Notice how we are providing multiple "argument selectors" now
     [
       (state) => state.favouriteTodoId,
       // Here we pull a reference to the other derived state selector
       (_, storeState) => storeState.todos.getById
     ],
     // We receive both our argument selectors and execute the received
     // derived state selector
     (todoId, getTodoById) => getTodoById(todoId)
    )
  }
});
```

### Final Notes

Some tips when using this feature:

1. Keep your "argument selectors" simple. They should simply return the bits of state that you care about. Do all the mapping and deep property accessing within the selector function itself - you will generally experience improved performance compared to doing complex selections within your "argument selectors".

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -47,6 +47,7 @@ type FilterActionTypes<T extends object> = Omit<
     | RegExp
     | Reducer<any, any>
     | Select<any, any>
+    | Derived<any, any, any, any, any>
     | Listen<any, any, any>
   >
 >;
@@ -59,7 +60,13 @@ type FilterStateTypes<T extends object> = Overwrite<
       Action<any, any> | Listen<any, any, any> | Thunk<any, any, any, any, any>
     >
   >,
-  Pick<T, KeysOfType<T, Select<any, any> | Reducer<any, any>>>
+  Pick<
+    T,
+    KeysOfType<
+      T,
+      Select<any, any> | Reducer<any, any> | Derived<any, any, any, any, any>
+    >
+  >
 >;
 
 /**
@@ -96,6 +103,8 @@ type RecursiveState<
     {
       [P in keyof RequiredModel]: RequiredModel[P] extends Select<any, infer R>
         ? R
+        : RequiredModel[P] extends Derived<any, any, any, any, any>
+        ? Selector<RequiredModel[P]>
         : RequiredModel[P] extends Reducer<infer R, any>
         ? R
         : RequiredModel[P] extends object
@@ -107,6 +116,8 @@ type RecursiveState<
     {
       [P in keyof OptionalModel]?: OptionalModel[P] extends Select<any, infer R>
         ? R
+        : OptionalModel[P] extends Derived<any, any, any, any, any>
+        ? Selector<OptionalModel[P]>
         : OptionalModel[P] extends Reducer<infer R, any>
         ? R
         : OptionalModel[P] extends object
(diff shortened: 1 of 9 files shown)


homura:
derived state looks good, but I think "runtime arguments" possibly make derived confused. See https://github.com/reduxjs/reselect#q-how-do-i-create-a-selector-that-takes-an-argument

PR #188, opened 2019-05-28
Adds Container and Local stores

### LINKED ISSUE #131: Local stores in components
I love how much easy-peasy simplifies global stores and I'd like to get same experience inside components. Very often I'd do HTTP request inside component and not share this data with other components. In past I used to solve this with simple MobX model, that would accept url, make request and set `.result` property on sucess or `.error` on failure. I would then instantiate this model in component and use it.

Easy-peasy models behave very much the same but I don't see a way to create local store, I can only use one provided by StoreProvider. Is it possible to create local store inside one component and have this component responding to store changes?

ctrlplusb:
Closes #131 

```javascript
import { createContainerStore, action } from 'easy-peasy';

const Counter = createContainerStore({
  count: 0,
  inc: action(state => {
    state.count += 1;
  }),
});

function CountDisplay() {
  const count = Counter.useState(state => state.count);
  const inc = Counter.useActions(actions => actions.inc);
  return (
    <>
      <div>{count}</div>
      <button onClick={inc} type="button">
        +
      </button>
    </>
  );
}

const app = (
  <Counter.Provider>
    <CountDisplay />
  </Counter.Provider>
);
```


dai4:
Just my 2 cents on this... I think you should rename `Counter.useState` to `Counter.useStore` to keep it consistent with the global store

PR #200, opened 2019-06-12
Derived receives the args directly I think

shoeman22:
I was doing my first selector and running into troubles with the array destructure args and when I changed it to just an arg list it seems to be working as intended.


shoeman22:
(I'm not sure why it shows a change on the last line about redux, but I can't seem to get it to go away for some reason)

ctrlplusb:
Nicely spotted @shoeman22 - I appreciate the help 👍

PR #204, opened 2019-06-15
Easy Peasy  competitive advantage

kigh143:
I tried to put out Easy Peasy to developers showing its pros. I have compared it more with setting up redux from scratch. This will help developers get a know what to expect when they use Easy Peasy.

### DIFF
--- a/website/docs/README.md
+++ b/website/docs/README.md
@@ -11,6 +11,21 @@ Easy Peasy provides you with an <strong>intuitive</strong> API to <strong>quickl
 
 <p>&nbsp;</p>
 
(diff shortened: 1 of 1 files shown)


ctrlplusb:
Hi @kigh143 👋

I'm really happy to see that you are digging Easy Peasy, that being said I am going to pass on this PR for now. I'd like to avoid this type of comparison for now, even though Easy Peasy is a direct alternative Redux binding for React. I still have much respect for React Redux and the team involved, and would prefer to advertise Easy Peasy based on its own merits.

Thanks anyways though. 

How has your Easy Peasy journey been going btw?

PR #228, opened 2019-06-29
Fix example for defining-computed-properties

allpro:
The docs example is missing the `state` object for `products`, i.e.: `products.reduce(...)` -> `state.products.reduce(...)`


ctrlplusb:
Thanks ♥️

PR #264, opened 2019-07-18
Adds meta support to actions

### LINKED ISSUE #241: Define a meta object on an action
I have a middleware that looks for actions that have a special property. In classic Redux I normally added a `meta` key to the action object, which holds information like this, e.g. `meta.canUndo` or `meta.batched`.
I thought maybe the second parameter of the `action()` helper could be used for something like this, but as of now it seems to only support the `listenTo` option.

Is there a way to have static meta information on an action right now? Or is it considerable to add this in the future?

Btw: With or without a meta object - easy-peasy is an amazing library! Thank you! :)

ctrlplusb:
Closes #241


ctrlplusb:
Closing this for now. I may revisit the feature at a later date.

PR #294, opened 2019-08-04
Fix useStoreState usage in quick-start

goncy:


ctrlplusb:
Thank you so much for the help. ❤️

Issue #309, opened 2019-09-04
Interop with an existing React Redux application


jaredmeakin:
I have a current React application that makes heavy use of combine reducers, and in some cases those reducers are nested. Is my only course of action to replace all instances of combined reducers with easy-peasy's reducer function?

Just trying to find some alternative options before I trudge down the refactor route. Thanks!

ctrlplusb:
Hey @jaredmeakin 

AFAIK you _should_ only need to replace the root level with easy peasy's `reducer`. Any nested `combineReducers` calls should be fine. 👍

PR #310, opened 2019-09-04
Update the `reducer` comment block in the Interop recipe.

jaredmeakin:
This relates to #309. Thanks for the clarification, and hope this is clear enough.


ctrlplusb:
Thanks so much for championing this change. 💜

PR #321, opened 2019-09-25
HTML suffix to store links within docs

jacekk:
Without those extensions, various places point to `https://easy-peasy.now.sh/docs/api/store` (no `.html` suffix), which results in 404.

Related docs: https://v1.vuepress.vuejs.org/guide/markdown.html#links


ctrlplusb:
I really appreciate you taking the time to fix these @jacekk. ❤️

Issue #227, opened 2019-06-29
Type instantiation is excessively deep


swarnakumar:
I just upgraded from 2.5.0 to 2.6.5.

Typescript is not compiling. 

My Code:
```const logged_in = useStoreState(store => store.auth.get_login_status());```

The message is 
```Type instantiation is excessively deep and possibly infinite```

This was working fine earlier.

**Comment by ctrlplusb** [comment]
Hi @swarnakumar 

What version of TS are u using?

**Comment by swarnakumar** [comment]
3.5.1

**Comment by swarnakumar** [comment]
There was an unused import of select in my ducks file where I was declaring the model.

The error vanished when I removed the import.

Closing the issue.

**Comment by swarnakumar** [comment]
I am seeing this issue still. And its intermittent.

The fix on my previous comment, ensured compilation happens. My project was bootstrapped using CRA.

 But VS Code screams quite a bit about this.

I am unable to figure out what's happening.

**Comment by strobox** [comment]
Strange error. Also disappeared after some manipulations (i.e. typescript upgrade)

**Comment by danielmahon** [comment]
Had this issue too, removing `node_modules` `yarn.lock` and reinstalling fixed it for now... using typescript@3.5.2 and easy-peasy@2.6.5

**Comment by danielmahon** [comment]
Downgrading didn't help, not sure why all of a sudden TS is throwing this error. Seems to go away if I get stricter with the types(which I suppose is what it wants).
```ts
import { Action, action } from 'easy-peasy';

export interface DataModel {
  screens: any[]; // changed from "any"
  files: any[]; // changed from "any"
  set: Action<DataModel, Partial<DataModel>>;
}

const data: DataModel = {
  screens: [],
  files: [],
  set: action((state, payload) => {
    return { ...state, ...payload };
  }),
};

export { data };
```

ctrlplusb:
Ah, interesting. I don't test for the any case. Also, are any of you using generic interfaces or types in your models?

ctrlplusb:
I am gonna do some more testing and work on the definitions 👍

danielmahon:
Popped up again in VSCode, though TS seems to be compiling fine (create-react-app project). Maybe there's some discrepancy in VSCode?

swarnakumar:
This got even weirder for me

Error pops up when:
```
const state = useStoreState(state => state.auth.something);
const action = useStoreActions(actions => state.auth.someAction);
```

Very weirdly, error stops when I reverse the two statements:
[...]

PR #322, opened 2019-09-28
reduce the number of mapped types used in state mapper

jmalins:
Working proposal on #227, not patch quality at this point.

My exact manifestation is:
 - Typescript 3.5.3 and 3.6.3
 - Problem only appears in the state mapper (actions, listeners are fine).
 - Reducing the recursion depth to 4 from the original 6 clears the error (suggesting ~10-12 levels of mapped types per recursion).
 - Error appears inconsistently between VSCode inline errors and CRA build process.

The typescript compiler has a hard instantiation depth limit of 50, as explained here:
https://github.com/microsoft/TypeScript/issues/31619
The TS team considers this "as designed". 

I haven't been able to create a standalone example to replicate the issue, but the patch below eliminates it in my environment. The intent is to reduce the number of type instantiations by replacing the elegant (but complex) `ts-toolbelt` based type mapper with a simpler alternative. I haven't done exhaustive model tests, but the code below seems to provide the same behavior as the original implementation, with a maximum of 3 levels of mapped types per recursion.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -215,13 +215,18 @@ type RecursiveState<
   Depth extends string
 > = Depth extends '6'
   ? Model
-  : O.Merge<
-      StateMapper<
-        O.Omit<O.Filter<Model, ActionTypes>, IndexSignatureKeysOfType<Model>>,
-        Depth
-      >,
-      O.Pick<Model, IndexSignatureKeysOfType<Model>>
-    >;
+  : {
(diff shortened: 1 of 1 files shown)


ctrlplusb:
Ah, this is amazing @jmalins - wow, I super appreciate you putting in the effort to optimise this type. Huge love. ❤️

Issue #306, opened 2019-09-01
State update on an unmounted component when using thunks


valstu:
I bumped into on issue where easy-peasy tries to make state update on a component which is unmounted. This happens when I'm using thunks to perform side effects which will trigger actions. Then I am using that state to define what components to render.

For the sake of simplicity I created this simple Codesandbox example.
https://codesandbox.io/s/festive-moore-9fjvz

I have a store which has `userId`. If the userId is undefined it will render `<Login />` component, otherwise it will render the `<Home />` component. Home component consumes the `userId` state and `logout` thunk. When you first click login button on Login component it renders the `<Home />` component as expected. Then if you click the logout button on the <Logout /> component the state is updated correctly and React renders the `<Login />` but it still apparently tries to update the state inside the Home component even though it is unmounted. You will notice the following React warning on the console.
```
Warning: Can't perform a React state update on an unmounted component.
```

This won't happen if I change the onClick on the logout button to `loggedOutSuccess` action instead of `logout` thunk.

jmalins:
I observed this as well when an asynchronous state update causes a component using `useStoreState` to be unmounted.

Wading through the stack trace, it appears that the state update in question is the call to `forceUpdate({})` (aka the s+1 simple reducer) in `checkMapState`.

TatisLois:
Were you able to resolve this warning?

jmalins:
Yes, I was, thanks for reminding me to post the results.

See PR #324 for fix and explanation.

ctrlplusb:
Fixed in v3.1.0

Thanks to @jmalins for the fix. 👍

AnsonH:
I suggest re-opening this issue since using **v5.2.0** will cause the same warning of "state update on an unmounted component":

https://user-images.githubusercontent.com/57580593/211547827-d8e127eb-5d88-49ff-b611-73ffa3fbbd32.mp4

Here's the Code Sandbox link (same code as the original sandbox, but with v5.2.0).

[![Edit Easy Peasy #306 in v5.2.0](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/easy-peasy-306-in-v5-2-0-tt1vtf?fontsize=14&hidenavigation=1&theme=dark)

I tried changing the version to v3.1.0 and the warning does NOT appear. It seems that some updates along the way re-introduced this bug.

jmyrland:
[...]

PR #324, opened 2019-10-01
Prevent state updates on unmounted components

jmalins:
Fixes issue (#306) that occurs under the following conditions:

 1. A piece of store state is modified.
 2. The state modification causes component(s) to unmount.
 3. One or more of the unmounted components is watching the modified state from 1 via a `useStoreState` hook.
 4. The `useStoreState` hook detects the watched state change and attempts to re-render the unmounted component.

I found this problem (along with the OP of #306) in authenticated user tracking: protected routes unmount without a logged in user, while the routes themselves use the user identity from the store.

The root cause here is that `redux` does not guarantee subscription changes during a dispatch cycle. The `useStoreState` hook is unsubscribed from the store as it is unmounted, but it still receives the state change notification due to a stale subscription.

Fix:
 - Use a ref to mark the hook as unmounted during unsubscribe.
 - Suppress the component update if unmounted.

### DIFF
--- a/src/hooks.js
+++ b/src/hooks.js
@@ -23,6 +23,7 @@ export function createStoreStateHook(Context) {
     const store = useContext(Context);
     const mapStateRef = useRef(mapState);
     const stateRef = useRef();
+    const mountedRef = useRef(true);
     const subscriptionMapStateError = useRef();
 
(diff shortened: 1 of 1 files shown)


ctrlplusb:
Another solid PR! Thanks so much ❤️

PR #329, opened 2019-10-11
Updates the link to the StoreConfig documentation.

OliDM:
Before this was pointing to a section that no longer exists on the current page.


ctrlplusb:
Good catch, thanks!

PR #343, opened 2019-10-30
Adds store persistence APIs

ctrlplusb:
In progress. Please feel free to beta test this. Feedback appreciated!

```
npm install easy-peasy@3.2.0-alpha.3
```

You can view the [updated website docs here](https://easy-peasy-next.now.sh/docs/api/persist.html)

**Todo**

- [x] Strategy to allow persist/hydration of multiple stores
- [x] More tests to gain 100% API coverage along with as many edge cases as I can guess
- [x] Rewrite the rehydrate/persist middleware to be API compatible with redux persist transforms
- [x] Docs, docs, docs!
- [x] Beta release
- [x] Clean code refactor!!! Need to completely rewrite the logic as took some leaps to get this working. But have a robust test suite now to allow me to refactor with.
- [x] Code optimisation rounds
- [x] TypeScript definitions
- [x] What about component and context stores? Need to test these and ensure compatibility with them is well documented.


allpro:
Wow, talk about timing! I just created a selective persistence/hydration system for store-models yesterday. I'll scrap that now and implement this instead.

I'll let you know how it goes.

allpro:
I'm persisting models individually because only a few values are persisted, and each model configures its own keys. I'm using a custom storage object that creates user-specific indexedDB databases.

I wish I could set the 'default storage object' in the createStore() config. Then each model would not have to set the storage option in its use of persist(). The default would replace the 'session' default, and could be overridden by any instance of persist, just as session can.

allpro:
When you estimate persist() will land in a non-beta release? 

I'm implementing this in a production app and the next version will ship in 5 weeks.

ctrlplusb:
Hey @allpro 

It is in V3.2.0 😄

I didn't leave it in beta too long as I was mostly matching the redux persist API.

Hope it works for ya.

allpro:
I implemented persist and it's working fine. Here is some feedback...

I am persisting user-specific data; using a custom service they creates a separate indexedDB store for each user. This works good for me.

However, for those who use local storage, caching data by-user is a problem. I suggest adding an option like "keyPrefix". This becomes part of the key-name, just as store-name is used. This allows a userId to be specified to segment the data in some models by-user, or some other segmentation like "role".

I delay creation of the store, and therefore rendering of the app, until the user data is fetched so we have the userId that identifies the correct db. The same issue would occur if using a keyPrefix option. A useful option would be to defer hydration until it is manually triggered, such as store.hydrate(). This allows the store to be created and the app to render before the userId, (or whatever), is received - or before login occurs.

As noted earlier, it would also be nice to be able to specify 'persist defaults' in the StoreConfig, to minimize repetition when using persist for individual models. In my case, I'd like to set the 'storage' and 'whitelist' options, because these are the same for most models.

ctrlplusb:
Hey @allpro 

I appreciate this feedback. ❤️

I will take time to digest it. 👍

ctrlplusb:
@allpro - thank you so much for the very generous contribution. ❤️

This will certainly help me with any future Easy Peasy plans. 👍

Issue #275, opened 2019-07-25
Return multiple actions from useStoreActions?


joelmoss:
So while the naming of `useStoreActions` would imply you can return multiple actions from a single call, is that safe/performant to do?

For example:

```javascript
  const { attachPaymentMethod, submit } = useStoreActions(x => ({
    attachPaymentMethod: x.attachPaymentMethod,
    submit: x.submit
  }))
```

I ask because it seems to be recommended that you return only a single value from `useStoreState`, so wandered if the same applies?

thx

ctrlplusb:
Completely fine to do this. 👍

joelmoss:
I figured as much, thx.

I gotta admit that it's _slightly_ annoying that when I need multiple state values from the store that I have to call `useStoreState` for each one...

```javascript
const therapistId = useStoreState(x => x.therapistId)
const submitButtonText = useStoreState(x => x.submitButtonText)
const view = useStoreState(x => x.view)
...
```

Becomes a little repetitive. With Redux, I liked being able to define a single `mapStateToProps` function to return a hash of the state that I need.

Now I know the reason for this - equality checking. But maybe some way to pass or define a custom equality function would be nifty...

```javascript
const shallow = () => { ... }
const { foo, bar } = useStoreState(state => ({ foo: state.foo, bar: state.bar }), shallow)
```

What do you think?

FYI: awesome lib! Keep up the good work 👌

allpro:
Since actions have no equality checking, I often import the entire model-actions object when I need to call _multiple_ actions from a component. I find the OOP syntax also helps readability in larger code blocks. So something like...

````javascript
const actions = useStoreActions(x => x.path.to.model);
...
actions.attachPaymentMethod(payload);
actions.submit(data);
// etc.
````

**Comment by ctrlplusb** [comment]
Hey @joelmoss 

Thanks 😀

> But maybe some way to pass or define a custom equality function would be nifty

That's a pretty decent feature request. I believe the react-redux hooks do something similar. I'll take this under consideration. 👍

**Comment by joelmoss** [comment]
I got the idea from Zustand https://github.com/react-spring/zustand#selecting-multiple-state-slices

**Comment by joelmoss** [comment]
And as actions have no equality check, how about this to make action fetching even more easy-peasy... 😁

```javascript
const [ attachPaymentMethod, submit ] = useStoreActions('attachPaymentMethod', 'submit')
```

I created a little wrapping hook to actually enable the above...

```javascript
[...]

PR #354, opened 2019-11-08
feat: Shallow equality option for useStoreState()

joelmoss:
Currently. useStoreState() performs a strict equality check (`oldState === newState`) on its state, which is great for most cases. But sometimes you need to work with more complex state and return something that will always fail a strict check.

This PR adds support for a shallow equality check as follows:

```javascript
const store = createStore({
  count: 1,
  firstName: null,
  lastName: null
});

// In your component
const { count, firstName } = useStoreState(
  state => ({
    count: state.count,
    firstName: state.firstName,
  }),
  { shallowEquality: true }, /* this is new, an is false by default */
);
```

In the above case, if either the `count` or `firstName` state vars change, the equality check with be true, and the component will re-render. And if any other state is changed; for example, the `lastName`, the equality check will be true, and the component will not re-render.

The `shallowEquality` option is false by default, thus maintaining existing functionality.

See #275


joelmoss:
@ctrlplusb anything I can do to push this along?

ctrlplusb:
Hi @joelmoss - shall endeavour to review this week. I have had a very busy past few weeks.

joelmoss:
No worries bud. Thx

Joel Moss
Develop with Style
DevelopWithStyle.com
07791 503309

https://www.linkedin.com/in/joelkmoss/
https://github.com/joelmoss
https://twitter.com/joelmoss

ctrlplusb:
@joelmoss I've had a look. Thanks so much for this. Will be an awesome addition.

I'd like to request that we change the API somewhat. Instead of the configuration object could we instead accept an equality function as the second argument to useStoreState. This would very much be in line with the Redux useSelector hook then.

We could include a default shallow equality fn that user's could leverage.

This provides a lot more flexibility going forward IMO.

joelmoss:
ok, yeah I like that. Leave it with me.

joelmoss:
By the way, I noticed that the `shallowequal` package is included, so we can use that as the officialy provided equality function.

However, I'm not seeing why that package is included and where it used.

joelmoss:
Actually, no matter. The `shallowequal` package accepts a customizer which we don't and won't ever use.

joelmoss:
done!

ctrlplusb:
Amazing thanks!

crissdev:
@joelmoss Thanks for this feature. If possible please squash your commits into one and update the docs/example found in this commit 1d29a67ed116c1e38fe4ff6cad406ded8e1094eb, since it no longer matches the final implementation. This might be very useful to @ctrlplusb if he decides to use that in the documentation.

joelmoss:
Done!

crissdev:
@joelmoss Cool. I think `index.d.ts` should also be updated to include the changes to `useStoreState` and the new `shallowEqual` function. Do you want to give that a try?

joelmoss:
Not familiar with TS, but will give it a go

ctrlplusb:
@joelmoss - if you aren't comfortable with doing the TS don't stress. Let me know and I will pick that up for ya.

joelmoss:
Yeah, sorry. Not familiar enough with TS to give this time. So any help would be appreciated, thx

Joel Moss
Develop with Style
DevelopWithStyle.com
07791 503309

https://www.linkedin.com/in/joelkmoss/
https://github.com/joelmoss
https://twitter.com/joelmoss

joelmoss:
@ctrlplusb Just had a thought about this PR in the context of computed properties; do they also have a strict equality check? If so, we should probably allow this new option with computed properties too.

Issue #347, opened 2019-11-06
How should we listen to a thunk successType and failType?


viniciusHagamenon:
Hi!

I'm trying to listen to `successType` and `failType` from an `thunk` but the `failType` is only dispatched if I don't catch the error from the original thunk but doing that I get uncaught exceptions as it's being re-thrown by the `thunk`.

Is there a way to trigger a `failType` without getting exceptions?

I have looked at the code and thought that maybe we could do something like this in the thunks.js:

```js
try {
      const result = references.dispatch(() => thunkHandler(payload));
      if (typeof result === 'object' && typeof result.then === 'function') {
        return result
          .then(resolved => {
            dispatchSuccess(resolved);
            return resolved;
          })
          .catch(err => {
            dispatchError(err);
            if (!isSomethingListeningToThisThunkFailType) { // <
                throw err;
            } // <
          });
      }
      dispatchSuccess(result);
      return result;
    } catch (err) {
      dispatchError(err);
      if (!isSomethingListeningToThisThunkFailType) { // <
          throw err;
      } // <
    }
```

Thanks for the lib by the way, it's awesome!

Schiavi:
This is somewhat related to #332, because the intention of this is to have a global error handler for thunks.

We are trying to avoid the Redux pattern of `ACTION` > `ACTION_SUCCESS` || `ACTION_FAIL` by using easy-peasy success/fail types. The problem is that the success always happens if we catch the thunk error inside the thunk, and we cannot let the thunk throw because it's not a silent fail.

ctrlplusb:
The v4 release will include some notes that addresses this issue. Stay tuned.

Issue #332, opened 2019-10-15
Handling try and catch logic, under thunks with a Higher Order Function.


yaldram:
- The Following is just a **Question** and not a **Bug**
- Hello all, I need some help regarding handling errors.
- Errors for 401 and 403 should be handled globally and the user should be redirected to Login Screen.
- For all other errors, I have my own Error Handlers.
-  My approach was to create a Higher Order Function like so
- **as we all know an easy-peasy thunk should always return a Promise in case of async functions and for listeners to work** : -

```javascript 
import { message } from "antd";

import { store } from "./index";

export const hoc = ({ loadingField, main, errorCallback, finallyBlock }) => (
  actions,
  payload
) => {
  if (loadingField) {
    actions.toggleLoading(loadingField);
  }
  return main(actions, payload)
    .then(_ => _)
    .catch(error => {
      if (!error) {
        message.error("Please Check Your Internet Connection");
        store.dispatch({
          type: "@action.users.logout"
        });

        return;
      }

      const { status, data } = error;

      if (status === 401 || status === 403) {
        message.error(data.message);
        store.dispatch({
          type: "@action.users.logout"
        });

        return;
      }

      if (errorCallback) {
        errorCallback(actions, error);
      } else {
        message.error(error.data.error);
      }
    })
    .finally(() => {
      if (finallyBlock) {
        finallyBlock(actions);
      } else if (loadingField) {
        actions.toggleLoading(loadingField);
      }
    });
}; 

```

- Second, my thunks are like below -

```javascript 
  editEmailConfigById: thunk(
    hof({
      main: async (actions, payload) => {
        actions.toggleLoading("editLoading");
        const { emailConfigId, updatedValues } = payload;
        const { data } = await editEmailConfigById(
          emailConfigId,
          updatedValues
        );
        actions.updateEmailConfigOnEdit(data.email_config);
        message.success("Updated Successfully");
      },
      errorCallback: error => {
      // My Custom Error Code
      },
      finallyBlock: actions => {
        actions.toggleLoading("editLoading");
      }
    })
  ),
```
- What I achieve by the `Higher Order Function` is I am able to reduce duplication while handling 
`401` and `403` errors and `No-Internet Connection`. 
- Most of my thunk errors I am just showing a simple message so I am just using `message.error()` in one place i.e. under the `HOC` reducing more duplication.

- `My Questions`
1. Is this the right way or `easy-peasy` way to handle try{}catch(){} in one place?
[...]

PR #362, opened 2019-11-15
feat: Returning falsey value from a thunk will not throw or reject

joelmoss:
There are times when I do not want to throw an error or reject a Promise. For example, I may have already handled a specific error scenario and want to do that within the thunk itself. I could handle the error within my components, but that means I would have to do so in every component that calls the thunk.

Currently, a thunk must return a promise or throw an error to ensure that all easy-peasy listeners are triggered. So if I catch any error in the thunk and not re-throw, the failType listeners will not be triggered.

This PR simply allows you to return a truthy/falsey value from your thunks, and supports its existing functionality.

If you return a truthy value, the thunk will be a success, and if you return a falsey value, then  the thunk will be a failure. And this results in the expected listeners being triggered.

As an example:

```javascript
users: thunk(async () => {
  try {
    await fetch('/users')
  } catch (error) {
    if (error.response.status === 404) {
      return false
    } else {
      throw error
    }
  }
})
```

The above will catch any 404 and return false but not throw or reject a promise. Otherwise, the original error is thrown.

### DIFF
--- a/src/__tests__/thunks.test.js
+++ b/src/__tests__/thunks.test.js
@@ -45,6 +45,106 @@ test('dispatches actions to represent a succeeded thunk', () => {
   expect(actualResult).toBe('did something');
 });
 
+test('returning falsey from thunk', () => {
+  // arrange
+  const model = {
+    foo: {
+      doSomething: thunk(() => {
+        return false;
+      }),
+    },
+  };
+  const trackActions = trackActionsMiddleware();
+  const store = createStore(model, { middleware: [trackActions] });
+  const payload = 'a payload';
+
+  // act
+  const res = store.getActions().foo.doSomething(payload);
+
+  // assert
+  expect(trackActions.actions).toEqual([
+    { type: '@thunk.foo.doSomething(start)', payload },
+    {
+      type: '@thunk.foo.doSomething(fail)',
+      payload,
+      error: false,
+    },
+    {
+      type: '@thunk.foo.doSomething',
+      payload,
+      error: false,
+    },
+  ]);
+
+  expect(res).toBe(false);
+});
+
+test('returning nothing or undefined from thunk', () => {
+  // arrange
+  const model = {
+    foo: {
+      doSomething: thunk(() => {}),
+    },
+  };
+  const trackActions = trackActionsMiddleware();
+  const store = createStore(model, { middleware: [trackActions] });
+  const payload = 'a payload';
+
+  // act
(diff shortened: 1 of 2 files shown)


joelmoss:
OK, so I just realised that this will completely fail if using async/await :(

Watch this space...

ctrlplusb:
@joelmoss - I may just catch you there. 

I'm not completely up for this idea, for two reasons:

1. It's a breaking change
2. I actual depend on being able to return values out of my thunks to the dispatching code. This is valid and quite useful in some cases. This proposal would muddy the semantics here.

I appreciate what you are trying to represent though and am wondering if we could look to an alternative approach. For e.g., what about an API that allows you to explicitly put the thunk into a failed state?

```javascript
users: thunk(async (_actions, _payload, { fail }) => {
  try {
    await fetch('/users')
  } catch (error) {
    if (error.response.status === 404) {
      fail(); // Force the thunk into a failure
    } else {
      throw error
    }
  }
})
```

We could even allow you to pass a reason/Error into the fail which would propagate:

```javascript
fail(new Error('404'));
```

It's just an idea for now, and one to play around with. The implications of this would need to be considered at a much deeper level.

Thoughts?

joelmoss:
Yeah well it appears my original idea of keeping it as organic as possible and allowing the return of true/false, along with maintaining backwards compat isn't gonna work. 

But I can see merit on being explicit with some kind of `fail()` call. let me think on this a little more and rework my PR.

joelmoss:
Again, in the interest of keeping this as organic and natural as possible, and to maintain backwards compat, how about we check for a `doNotThrow` property on any `Error` that is returned, and not throw it if the property is true...

```javascript
users: thunk(async (_actions, _payload, { fail }) => {
  try {
    await fetch('/users')
  } catch (error) {
    if (error.response.status === 404) {
      error.doNotThrow = true
    }

    throw error
  }
})
```

So if `error.doNotThrow === true`, the error is not actually thrown, but still returned.

Thoughts?

viniciusHagamenon:
I like the idea of `fail(props)` better as it would allow you to pass you error and/or payload to the potential `thunkOn` failType listeners.

This idea helps solve #347 and #332 allowing to create this logic around to better handle thunk exceptions and async flow using `startType`, `successType`, `failType`.

ctrlplusb:
Personally i prefer the `fail` API approach too - it feels more explicit and flexible.

ctrlplusb:
I am closing this PR, as I have a solution for this coming soon. 💜

PR #366, opened 2019-11-22
update docs for store

crissdev:
I don't see any reason to include `required` here - maybe a copy-paste remnant?


ctrlplusb:
Yay! I love that you are helping with the docs. Thanks so much for these PRs. ❤️

ctrlplusb:
I completely agree with you about the *required*/*optional* being unnecessary in these cases.

PR #368, opened 2019-11-22
update docs for computed

crissdev:
Use `stateResolvers` instead of `stateSelectors` to avoid confusion.

---
One thing I don't know is to what `deprecated selector` is referring to in the following section  
https://github.com/crissdev/easy-peasy/blob/patch-3/website/docs/docs/api/computed.md#supporting-runtime-arguments

### DIFF
--- a/website/docs/docs/api/computed.md
+++ b/website/docs/docs/api/computed.md
@@ -10,11 +10,11 @@ isLoggedIn: computed(state => state.user != null)
 
 ## Arguments
 
-- `stateSelectors` (Array\<Function\>, *optional*)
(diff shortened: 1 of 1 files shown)


ctrlplusb:
> One thing I don't know is to what deprecated selector is referring to in the following section
> https://github.com/crissdev/easy-peasy/blob/patch-3/website/docs/docs/api/computed.md#supporting-runtime-arguments

It was an old API which was removed. We should probably remove mention of this from the docs. I think it was a copy and paste job from my original docs I included in the PR for `computed`

PR #392, opened 2019-12-28
Added runtime injections support for context/local stores

kamikazePT:
First of all, great library!
Been using it and spreading it to friends and co-workers :)

One "feature" that always seemed missing to me, was the ability to inject dependencies at runtime for thunk actions when using context store and local store.

For example, imagine that I want access to an i18n instance ( for example with the useTranslation hook from https://react.i18next.com/latest/usetranslation-hook#usetranslation-params ) on a thunk and get the current language from it, instead of polluting every call with an extra argument... we could inject them runtime :)

one issue that I haven't tackled was keeping the state when new injections are set or changed ( hence re-creating the store )

But before I tackle it, I wanted to share the idea of this possible feature and get some feedback if It's good or bad, if It has any issues I'm not thinking about... 

PS: This was just a quick demo to get feedback, another thing is that I assumed only objects as injections, if the idea is accepted as a feature, I'll work on having things more clean and prepared for more use cases


kamikazePT:
@ctrlplusb 

assuming the heart emoji was an "aprooval", I've added the state persistance when injected dependencies are updated :)

added tests for that case on both local and context store

although on the context store test I add to force a render after the click for the child component to update... Not sure if it's some limitation with testing-library or perhaps there's some issue on the context store hooks?

Take a look at it please :)

Also I decided to assume that the "injections" prop on the store was always an object, because using it as something else seemed weird...

src/create-context-store.js
```diff
@@ -9,18 +16,32 @@ import {
 } from './hooks';
 import createStore from './create-store';
 
-export default function createContextStore(model, config) {
+export default function createContextStore(model, config = {}) {
   const StoreContext = createContext();
 
-  function Provider({ children, initialData }) {
+  function Provider({ children, initialData, ...injections }) {
```
ctrlplusb: I have given this a bit more thought and wonder if we should instead allow you to provide a configuration object instead of just injections. If a configuration object was defined in the `createContextStore` call then the prop via the `Provider` would override it. 

This would provide with a great deal of flexibility to have build/runtime defined configurations.

Consumers would also be able to do their own merge strategies on the `Provider` prop config too should they like.

i.e.

```typescript
<MyStore.Provider config={{ injections: { foo: 'bar' } }}>
```

src/create-context-store.js
```diff
@@ -9,18 +16,32 @@ import {
...
+    const initialDataRef = useRef(initialData);
+    const previousStateRef = useRef();
+
     const store = useMemo(
```
ctrlplusb: Based on my other comment regarding being able to provide a custom configuration via the `Provider` prop I think we should refactor this to:

```typescript
function Provider({ children, initialData, config: runtimeConfig }) {
  const store = useMemo(
    () =>
      createStore(
        typeof model === 'function' ? model(initialData) : model,
        runtimeConfig || config,
      ),
    []
  );
```

I think we should also make it a single opportunity, i.e. not allow for new configs to be provided during runtime. This would preserve the simplicity around the `useMemo` dependencies and maintain current behaviour of the store.

Are there genuine use-cases where you want to dynamically reconfigure a store, its injections etc?

src/create-component-store.js
```diff
@@ -1,23 +1,32 @@
...
 
 /**
  * Some good references on the topic of reinitialisation:
  * - https://github.com/facebook/react/issues/14830
  */
 
-export default function createComponentStore(model, config) {
-  return function useLocalStore(initialData) {
+export default function createComponentStore(model, config = {}) {
+  return function useLocalStore(initialData, injections = {}) {
```
ctrlplusb: The same comments apply here as I made on the context store. We should consider refactoring this to allow a "runtime" config to be provided.

src/create-context-store.js
kamikazePT: being able to pass the configuration runtime, could be useful, although some of the options dont make much sense to be defined, for example "disableImmer"

when you create the contextStore or localStore you define your actions based on the fact that Immer will be on or off, so It doesn't make sense to change It's value runtime

devTools doesn't seem to make sense either, because it could be set conditionally on the store creation based on env variables, having toggles runtime for it, doesn't seem that useful

name of the store could make sense if we we're reusing the same "context/local store" on many levels or in parallel ( same level on the tree ) and we want to distinguish them on dev tools... That one really seems to add some value

perhaps instead of passing a [StoreConfig](https://easy-peasy.now.sh/docs/api/store-config.html), perhaps just a sub-set of it?

src/create-context-store.js
kamikazePT: The injections I really think they should have the flexibility to be added/removed/modified, instead of one shot only like the rest.

Perhaps I'm over-engineering here, but if I'm using a localization library like i18next, and I have a context store that holds a json schema to build a form, with a thunk to fetch the schema

The API needs to get a query param which is the "language", in order for the schema to return with translated labels.

I currently have 2 options for that:
- add a model holding the current language and a setLanguage action to the store creation
and have a listener on the i18n instance to be notified for language changes, calling setLanguage to update the store.
 
PROS:
things were coupled together

CONS:
for every store that needed the language, i would need to have the i18next listener syncing to every store, and every store would share the same model for the language (too much repition)


- add the language as a parameter of the actions
PROPS
no need to sync

CONS
I would need to fetch the current language ( using I18next hooks) before calling every action and pollute the actions definition with an extra parameter

both these approaches would start o add lots of extra coding for every new dependency I needed...

The active language could just be injected runtime on the store provider and be accessible on the thunks

Or maybe I could want to inject a storage (sessionStorage or localStorage or some memory implementation) to fetch data on a thunk without making it tightly coupled to a storage implementation on the creation, perhaps I would re-reuse the same context store on 2 places, one would be injected with sessionStorage and the other with local storage, or maybe the context store could be shared between different apps, both with different storages (one could use localStorage and the other using an API wrapped on the storage API, or cookies wrapped on the storage API )

src/create-component-store.js
kamikazePT: same answers here :P

src/create-context-store.js
ctrlplusb: Hmmm, I get what you are trying to say. I think my concern is that it will be really tricky to apply a strategy by which to calculate if we should or should not be creating a new store. Some user's may not be providing immutable instances to the provider, in which case the store would get recreated for every render of the component housing the provider.

It feels like both options have inherent complexity that will be shifted to the user. Perhaps we can introduce a new prop to address this need from another angle:

```typescript
<MyStore.Provider 
   config={customConfig}
   shouldRecreateStore={(prevProps, nextProps) => {
     return prevProps != null
         && prevProps.config.injections.i18n !== nextProps.config.injections.i18n;
   }}
>
  <App />
</MyStore.Provider>
```

The default behaviour would be that the store is never recreated, but then users could provide this function and do the check themselves to force the store to be recreated under the custom condition they provide.  This would be far more explicit, and I love that it is opt-in.

Thoughts?

src/create-context-store.js
ctrlplusb: Yep, good point. Happy to document that the API will only support a subset of `StoreConfig`. 👍

src/create-context-store.js
kamikazePT: I belive that instead of "prevProps" and "nextProps" it should be "prevConfig" and "currConfig" because there won't be any other props to compare, even initialData will be replaced by the initialState of the StoreConfig, right?

"Some user's may not be providing immutable instances to the provider" seems like a valid argument

the default behaviour for shouldRecreateStore could be to iterate every prop on config and flatten injections for comparison

How would you refactor the useMemo that holds the store creation to toggle it's recalculation?

I'm thinking of something like...

`
    const previousConfigRef = useRef(config);
    const [toggledForceRender, setForceRecreate] = useState(false);
    
    useEffect(() => {
     if(shouldRecreateStore(previousConfigRef.current, config){
      setForceRecreate(!toggledForceRender);
     }
     previousConfigRef.current = config;
    }) 

    const store = useMemo(
      () =>
        createStore(
          typeof model === 'function'
            ? model(previousStateRef.current || initialDataRef.current)
            : model,
          produce(config, draft => {
            draft.injections = { ...draft.injections, ...injections };
          }),
        ),
      toggledForceRender,
    );
`

ctrlplusb: > I belive that instead of "prevProps" and "nextProps" it should be "prevConfig" and "currConfig" because there won't be any other props to compare, even initialData will be replaced by the initialState of the StoreConfig, right?

Yep, for now, but focussing on props rather than the config prop itself provides us with a nice future proof API. We _may_ in the future add additional props to the provider. 👍

> the default behaviour for shouldRecreateStore could be to iterate every prop on config and flatten injections for comparison

Firstly, let's rename it to `shouldCreateStore`.

I would prefer the default behaviour, for when no `shouldCreateStore` is provided, to only create the store for the first render. This would then maintain current behaviour.

> How would you refactor the useMemo that holds the store creation to toggle it's recalculation?

Quite dirty, but my initial thoughts would be...

```typescript
import React, { useLayoutEffect, useRef } from 'react';

export default function createContextStore(model, config = {}) {
  const StoreContext = createContext();

  function Provider(props) {
    const { children, initialData, config: runtimeConfig, shouldCreateStore } = props;

    const prevPropsRef = useRef();
    const nextPropsRef = useRef();
    const storeRef = useRef();

    if (nextPropsRef.current != null) {
      prevPropsRef.current = nextPropsRef.current; 
    }
    nextPropsRef.current = props;

    if (storeRef.current == null 
      || (shouldCreateStore && shouldCreateStore(prevPropsRef.current, nextPropsRef.current))) {
      storeRef.current = createStore(
        typeof model === 'function' ? model(initialData) : model,
        runtimeConfig || config,
      );
    }

    return (
      <StoreContext.Provider value={storeRef.current}>{children}</StoreContext.Provider>
    );
  }

  function useStore() {
    return useContext(StoreContext);
  }

  return {
    Provider,
    useStore,
    useStoreState: createStoreStateHook(StoreContext),
    useStoreActions: createStoreActionsHook(StoreContext),
    useStoreDispatch: createStoreDispatchHook(StoreContext),
    useStoreRehydrated: createStoreRehydratedHook(StoreContext),
  };
}
```

kamikazePT: @ctrlplusb 
hey, I managed to understand the re-rendering issue on one of the tests, there was a bug on the createStateHooks,the forceRender wasn't working for store recreations, managed to fix it :)

also I've implemented this new approach of the shouldRecreateStore and added a configMergeStrategy, because runtimeConfig should be a subset of config, so you cant "replace" one for the other, they should be merged, by default it deep merges it, although it should probably validate if it isnt merging unwanted options like "devtool" or "disableImmer"

by default the shouldRecreateStore watches the injections (just for testing purposes of this feature...) that can be changed to a no-op

although the "no-op" seems misleading when you are passing a brand new config and the store remains the same old one

this new API also seems way more verbose than assuming props where injections, but this way its really more flexible...

what are your thoughts so far?

PS: I haven't updated the componentStore

kamikazePT:
@ctrlplusb 
going to make a fork from scratch and put this on a feature branch instead of my master... because i want to work on other stuff, i'll reopen eventually the PR with some changes I have in mind

PR #399, opened 2020-01-14
Remove note about delay

nickmeldrum:
At this point in the tutorial we aren't using the basket service that has this delay, so this note is unnecessary, potentially confusing?


ctrlplusb:
Woot! Contributing now too. 💜

Total convert. 😝

nickmeldrum:
😎

Issue #410, opened 2020-01-21
[Proposal] Lazy models


ctrlplusb:
The `addModel` and `removeModel` API doesn't feel as integrated as the rest of the APIs. This proposal will detail an inline approach of declaring your code-split/lazy-loaded models.

Details coming soon.

kamikazePT:
My proposal for this would be "wrapping" those 2 functions as in:

instead of 
`
import { StoreProvider } from 'easy-peasy'`

we would do

```
import { Store } from 'easy-peasy'

<Store.Provider store={store}>
...
</Store.Provider>
```

and on a code-splitted module deep in the react tree....

```
import { Store } from 'easy-peasy'

<Store.Model model={model}>
...
</Store.Model>
```

where model would be something as

```
{
 [modelName] : { 
  modelDefinition
 }
}
```

Store.Model would have a useEffect hook that runs only once, calling "store.addModel" for each key of "model" prop, and on the cleanup of the useEffect, calling "store.removeModel" for each key of the "model" prop :)

if you dont like "Store" with Provider and Model components

we can keep StoreProvider and add StoreModel

@ctrlplusb what you think?

ctrlplusb:
Very interesting API @kamikazePT 💜

dan-dr:
Microsoft did something similar to @kamikazePT 's suggestion for redux: https://github.com/microsoft/redux-dynamic-modules

There's also redux-injectors that seems like it will be easy enough to port
https://github.com/react-boilerplate/redux-injectors

ctrlplusb:
So here is the API I imagined.

**Defining a lazy model**

```javascript
import { createStore, lazy } from 'easy-peasy';
//                     👆
import todosModel from './models/todo';

const store = createStore({
  todos: todosModel,
  // We define a lazy model similar to how you define lazy react components 😊
  //          👇
  products: lazy(() => import('./products-model')),
});
```

**Loading a lazy model**

```javascript
import { LoadLazyModel } from 'easy-peasy';

function ProductsRoutes() {
  return (
    <LoadLazyModel target={model => model.products}>
      <Routes>
        {/* ... */}
      </Routes>
    </LoadLazyModel>
  )
}
```

**Using a loaded lazy model**

Once it is loaded you can use it as normal.

```javascript
import { useStoreState } from 'easy-peasy';

function ProductDetails({ id }) {
  const product = useStoreState(state => state.products.byId[id], [id]);
  return (
    <>
[...]

PR #412, opened 2020-01-23
feat: Lazy models [WIP]

kamikazePT:
following the approach detailed [https://github.com/ctrlplusb/easy-peasy/issues/410](url)

**Added**
 - useStoreModel hook
 - Store.Model wrapper component for a declarative approach

**TODO**
Tests

**Possible issues**
the usage of this new API on context stores might give some headaches if my proposal for runtime injections for context stores gets eventually merged, need to think about that...

@ctrlplusb 
 - Should this be available on context stores aswell?
 - Backwards compatibility with V3 so no need to wait for 4.0 :)

### DIFF
--- a/src/create-context-store.js
+++ b/src/create-context-store.js
@@ -1,10 +1,12 @@
 /* eslint-disable react/prop-types */
 
-import React, { createContext, useContext, useMemo } from 'react';
+import React, { createContext, useMemo } from 'react';
 import {
   createStoreActionsHook,
   createStoreDispatchHook,
   createStoreStateHook,
+  createStoreHook,
+  createStoreModelHook,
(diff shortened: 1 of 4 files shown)


kamikazePT:
@ctrlplusb 
hey again, going to pick this up soon as well, anything you would like to discuss regarding the proposal?

ctrlplusb:
Let's wait till after the next major release before tackling this.

I believe it can be bolted on as a minor version. I need the dust to settle so we can get through this next big release.

ctrlplusb:
I'm closing this, as I want to avoid adding additional complexity to this library right now as I don't have the capacity to maintain it regularly.

PR #416, opened 2020-01-29
Update create-your-store.md code example

saharbit:
While the example is clear for any experienced javascript developer, we want to make it accessible for newcomers. The examples given should work even when copy-pasted


ctrlplusb:
Nice catch @saharbit!

Thank you for this contribution. 💜

PR #433, opened 2020-02-18
fix localStorage error when in iframe

chrtze:
When embedding an easy-peasy app in iframe some users are getting errors because of their 3rd party cookie settings (https://stackoverflow.com/questions/30481516/iframe-in-chrome-error-failed-to-read-localstorage-from-window-access-deni). This can be fixed by wrapping the localStorage and sessionStorage getter into a try catch block.

### DIFF
--- a/src/persistence.js
+++ b/src/persistence.js
@@ -8,15 +8,22 @@ const noopStorage = {
   removeItem: () => undefined,
 };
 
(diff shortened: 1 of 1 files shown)


ctrlplusb:
Ah, nice one! 💜

PR #453, opened 2020-03-18
fix typo: website -> docs -> quick-start.md

hualu00:
url in website: https://easy-peasy.now.sh/docs/quick-start.html#action-thunk-listeners

should be 'state.logs.**push**(`Added product ${target.payload} to basket`)'  

it seems that 'push' is lost accidentally.


ctrlplusb:
Hi @hualu00 👋

Thanks so much for this contribution. Means so much to us keeping the quality of the documentation high.

💜

PR #455, opened 2020-03-19
fixes computed error when get() returns undefined

### LINKED ISSUE #400: addModel and computed props
I'm getting an error when I try to use computed props with addModel.

`TypeError: Object.defineProperty called on non-object`

```
store.addModel("session", {
  user: {
    userId: 5,
    firstName: "Felisha",
    lastName: "Blaine",
    email: "DylanThibodeaux@example.com",
    avatarUrl: "http://esionve.tw/asthiin/eraiten/shoesit/ititishad.gif"
  },

  id: computed(state => state.user.userId)
});
```

If I comment out the computed prop, everything works as normal. If I add a breakpoint to where the error is thrown I do see that the prop it's trying to run define on is  "undefined". I did some digging through the call stack but can't quite figure out whats going on here. I think it has something to do with this part of  easy-peasy lib.

```
function get(path, target) {
  return path.reduce(function (acc, cur) {
    return isPlainObject(acc) ? acc[cur] : undefined;
  }, target);
}
```
if stop execution and look at `acc`, I do not see my new "session" model in the object, so the value is undefined.

Appreciate any help!

**Edit**
I added some logging to the compute to make sure I wasn't missing something with the state.
```
  id: computed((state) => {
    console.log('id computed', state);
    return state.user.userId;
  })
```
Now i'm more confused as it seems like within the callback the computed is calculated properly?!
[https://take.ms/JzXRt](https://take.ms/JzXRt)

jchamb:
Fixes #400 
@ctrlplusb I was never able to truly track down why the next prop in get() would be missing the newly added model definition, but the reduce would return an undefined which would throw the error. Simply validating existence after the get() made the error go away and still performed the correct binding of the computed.

In the repo that has this issue, I am using a store provided through a package dependency and it does a number of addModel calls directly after one another, but I was never able to fully reproduce the issue consistently.

### DIFF
--- a/src/create-reducer.js
+++ b/src/create-reducer.js
@@ -48,7 +48,7 @@ export default function createReducer(
     if (state !== next) {
       computedProperties.forEach(({ parentPath, bindComputedProperty }) => {
         const prop = get(parentPath, next);
-        if (prop) bindComputedProperty(prop);
+        if (prop != null) bindComputedProperty(prop);
       });
     }
     return next;
--- a/src/create-reducer.js
+++ b/src/create-reducer.js
@@ -47,7 +47,8 @@ export default function createReducer(
         : stateAfterActions;
     if (state !== next) {
       computedProperties.forEach(({ parentPath, bindComputedProperty }) => {
-        bindComputedProperty(get(parentPath, next));
+        const prop = get(parentPath, next);
+        if (prop) bindComputedProperty(prop);
       });
     }
     return next;
--- a/src/__tests__/dynamic-store.test.js
+++ b/src/__tests__/dynamic-store.test.js
@@ -1,4 +1,4 @@
-import { action, createStore } from '../index';
+import { action, createStore, computed } from '../index';
 
 test('addModel', () => {
   // arrange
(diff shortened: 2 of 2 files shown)


ctrlplusb:
Hey @jchamb;

Thanks for this. This seems likely a reasonable temporary fix. I'll make a note to do some more digging into why the `undefined` case was occurring in the first place.

Appreciate your help!

👍

src/create-reducer.js
```diff
@@ -47,7 +47,8 @@ export default function createReducer(
         : stateAfterActions;
     if (state !== next) {
       computedProperties.forEach(({ parentPath, bindComputedProperty }) => {
-        bindComputedProperty(get(parentPath, next));
+        const prop = get(parentPath, next);
+        if (prop) bindComputedProperty(prop);
```
ctrlplusb: Actually I have approved this PR already, but we should avoid implicit boolean conversion. I would change this to:

```typescript
if (prop != null) bindComputedProperty(prop);
```

jchamb:
@ctrlplusb once I get some extra free time i'll create some repos I can share that mimic my split setup. I tried reproducing directly within a single repo and never could, so think this might be an issue with having my main store coming in as part of a dependancy.

PR #458, opened 2020-03-22
Adds useLocalStore hook, deprecates createComponentStore

ctrlplusb:
This PR deprecates `createComponentStore`. The API for `createComponentStore` was a bit verbose and limited. It will be removed in the next major release.

I plan to release non breaking changes for a while still, so no stress.

# useLocalStore

Allows you to create a store to represent the state for an individual React component. This is essentially an alternative to the official `useState` and `useReducer` hooks provided by React for component-level state.

```javascript
function MyCounter() {
    count: 0,
    increment: action(_state => {
      _state.count += 1;
    })
  }));

  return (
    <div>
      {state.count}
    </div>
  );
}
```

## Arguments

  - `modelCreater` (() => Object, required)

    A function that should return the model representing your store.

    ```javascript
    createComponentStore({
      count: 0,
      increment: action(state => {
        state.count += 1;
      })
    });
    ```

  - `dependencies` (Array, optional, default=[])

    A list of dependencies, similar to the official React hooks. If any of the values change then the store will be recreated. Especially helpful if your store depends on an external value, such as a prop.

    By default an empty array will be used, meaning the store will not be recreated.

  - `storeConfig` (Object, optional)

    Provides custom configuration options for your store. Please see the [StoreConfig](/docs/api/store-config.html) API documentation for a full list of configuration options.

    ```javascript
    useLocalStore(() => model, [], { middleware: [loggerMiddleware] });
    ```

## Returns

The hook returns a positional array;

```typescript
const [state, actions, store] = useLocalStore(/*...*/);
```

- `state`

  The state of your store.

- `actions`

  The actions of your store.

- `store`

  The actual store instance - for advanced cases.

## Example

This example shows how you can use the hook within your component.

```javascript
import { useLocalStore } from 'easy-peasy';

function MyCounter() {
    count: 0,
    inc: action(_state => {
      _state.count += 1;
    })
  }));
  return (
    <>
      <div>{state.count}</div>
      <button onClick={() => actions.inc()} type="button"> + </button>
    </>
  );
}
```

## Reinitialising your model based on incoming props

There may be cases in which you may want to reinitialize your store based on some incoming props. The store hook will only initialize once for each use. To get around this simply define the dependency array, providing a list of props, that when they change will cause the store to be recreated.

```javascript
import { useLocalStore, action } from 'easy-peasy';

function EditProduct({ id, name }) {
  const [state, actions] = useLocalStore(
    () => ({
      id,
      name,
      setName: action((_state, payload) => {
        _state.name = payload;
      })
    }),
    [id, name] // 👈 if any of these change the store is recreated
  );

  return (
    <>
      <div>Product {state.id}</div>
      <input
        onChange={e => actions.setName(e.target.value)}
        type="text"
        value={state.name}
      />
    </>
  );
}
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -799,6 +799,15 @@ export function createComponentStore<
   config?: StoreConfig,
 ): UseLocalStore<StoreModel, InitialData>;
 
+export function useLocalStore<
+  StoreModel extends object = {},
+  StoreConfig extends EasyPeasyConfig<any, any> = EasyPeasyConfig<any, any>
+>(
+  modelCreator: () => StoreModel,
+  dependencies?: any[],
+  storeConfig?: StoreConfig,
+): [State<StoreModel>, Actions<StoreModel>, Store<StoreModel, StoreConfig>];
+
 // #endregion
 
 // #region Persist
--- a/package.json
+++ b/package.json
@@ -52,7 +52,8 @@
     "redux": "^4.0.5",
     "redux-thunk": "^2.3.0",
     "symbol-observable": "^1.2.0",
-    "ts-toolbelt": "^6.1.6"
+    "ts-toolbelt": "^6.1.6",
+    "use-memo-one": "^1.1.1"
   },
   "devDependencies": {
     "@babel/cli": "^7.8.3",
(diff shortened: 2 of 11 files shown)


onionhammer:
How do you combine `useLocalStore` with `persist` ?

GollyJer:
@ctrlplusb I'm very much looking forward to this functionality. Is there a release on the horizon?

ctrlplusb:
I'll be cutting v4 this coming weekend. 😎

GollyJer:
🎉

PR #459, opened 2020-03-22
v3.4.0

### LINKED ISSUE #300: Incorrect typescript type when using a function to generate Models with generic types
(duplicate of https://github.com/ctrlplusb/easy-peasy/issues/189, but it seems to have broken again)

I am trying to make a model factory function to generate simple models with just a value and a setter action, where the value is of a generic type. When doing this the action's state isn't correctly typed.

example: 
```tsx
import { Action, action } from "easy-peasy";

interface SimpleModel<T> {
  value: T;
  set: Action<SimpleModel<T>, T>;
}

function makeSimpleModel<T>(initialValue: T): SimpleModel<T> {
  return {
    value: initialValue,
    set: action((state, payload) => {
      state.value = payload;
    })
  };
}

```

...and heres a quick code sandbox https://codesandbox.io/s/sparkling-cookies-4tx0m (doesn't do anything functional, just to demonstrate the red squiggly)

This generates a compiler error:
```
Property 'value' does not exist on type '_Merge<StateMapper<_Pick<_Pick<ListModel<T>, ({ 1: never; 0: "value"; }[Extends<T, ActionTypes>] & "value") | ({ 1: never; 0: "value"; }[Extends<T, ActionTypes>] & "set") | ({ 1: never; 0: "value"; }[Extends<...>] & "add") | ({ ...; }[Extends<...>] & "remove")>, (Exclude<...> & ... 1 more ... & "value") | ... 14 mor...'.ts(2339)
```

easy-peasy version: 3.0.2 (most recent)
typescript version: 3.5.3 (most recent)

### LINKED ISSUE #355: Improve persist rehydration logic to be more kind to an evolving model
The strategy  needs to improve to ensure that the expected model schema is still adhered to.

### LINKED ISSUE #359: Typescript decorators
Easy-peasy provides some nice abstractions over redux boilerplate but I think Typescript usage can be varied with the help of decorators. During my current project, I created a basic decorator implementation which simply collects some metada from class models and creates an easy-peasy store. I have yet to find any opportunity to use easy-peasy extensively since it takes a rather humble part of my project, still it did great for simple use cases with decorators. A full implementation -if possible- could be very useful.

Edit: Of course there are might be pitfalls in typings because my implementation is rather shallow.

example.store.ts
```
import { Computed, createStore, createTypedHooks, Model, Thunk } from "./easy-peasy-decorators";

@Model("counter")
export class CounterModel {
    public val = 0;

    public get nextCount(): Computed<number> {
        return this.val + 1;
    }

    @Thunk
    public async incrementThunk() {
        this.increment();
    }

    public increment() {
        ++this.val;
    }

    public reset() {
        this.val = 0;
    }
}

interface IStoreModel {
    counter: CounterModel;
}

const store = createStore<IStoreModel>();
```

easy-peasy.decorators.ts
```
import * as easyPeasy from "easy-peasy";

const model: any = {};
const instances: Record<string, any> = {};
const listeners: Record<string, Record<string, easyPeasy.TargetResolver<any, any>>> = {};
const thunks: Record<string, string[]> = {};
let store: easyPeasy.Store;

export function Model(modelName: string) {
    return (ctor: any) => {
        instances[modelName] = new ctor();
        model[modelName] = {};
        listeners[modelName] = {};
        thunks[modelName] = [];

        addState(modelName);
        addActionsAndComputeds(ctor.name, modelName);
    };
}

function addState(modelName: string) {
    const properties = Object.keys(instances[modelName]);

    properties.forEach(property => {
        const initialVal = instances[modelName][property];

        model[modelName][property] = initialVal;
    });
}

function addActionsAndComputeds(ctorName: string, modelName: string) {
    const prototype = instances[modelName].constructor.prototype;
    const descriptors = Object.getOwnPropertyDescriptors(prototype as object);

    Object.entries(descriptors)
        .filter(([methodName, desc]) => methodName !== "constructor")
        .forEach(([methodName, desc]) => {
            const { value, get } = desc;

            if (listeners[ctorName]?.[methodName]) {
                model[modelName][methodName] = easyPeasy.actionOn(
                    listeners[ctorName][methodName],
                    (state, target) => {
                        value.call(state, target);
                    },
                );
            } else if (thunks[ctorName] && thunks[ctorName].includes(methodName)) {
                model[modelName][methodName] = easyPeasy.thunk((actions, payload, { getState }) => {
                    value.call({ ...getState(), ...actions }, payload);
                });
            } else if (value) {
                model[modelName][methodName] = easyPeasy.action((state, payload) => {
                    value.call(state, payload);
                });
            } else if (get) {
                model[modelName][methodName] = easyPeasy.computed(state => {
                    return get.call(state);
                });
            }
        });
}

export function Listener<Model extends object, StoreModel extends object = {}>(
    actionFn: easyPeasy.TargetResolver<ToStoreType<Model>, ToStoreType<StoreModel>>,
) {
    return (ctor: any, methodName: string) => {
        listeners[ctor.constructor.name] = listeners[ctor.constructor.name] || {};
        listeners[ctor.constructor.name][methodName] = actionFn;
    };
}

export function Thunk(ctor: any, methodName: string) {
    thunks[ctor.constructor.name] = thunks[ctor.constructor.name] || [];
    thunks[ctor.constructor.name].push(methodName);
}

type ToStoreType<T extends object> = {
    [P in keyof T]: "computed" extends keyof T[P]
        ? T[P] extends Computed<infer U>
            ? U
            : T[P]
        : T[P] extends (...args: any[]) => any
        ? easyPeasy.Action<T, Parameters<T[P]>[0]>
        : T[P] extends object
        ? ToStoreType<T[P]>
        : T[P];
};

export function createStore<T extends object>() {
    store = easyPeasy.createStore<any>(model);

    return store as easyPeasy.Store<ToStoreType<T>>;
}

export function createTypedHooks<Model extends object>(): {
    useStoreActions: <Result>(
        mapActions: (actions: easyPeasy.Actions<ToStoreType<Model>>) => Result,
    ) => Result;
    useStoreDispatch: () => easyPeasy.Dispatch<ToStoreType<Model>>;
    useStoreState: <Result>(
        mapState: (state: ToStoreType<Model>) => Result,
        dependencies?: any[],
    ) => Result;
} {
    const hooks = easyPeasy.createTypedHooks<any>();

    return hooks as any;
}

export type Computed<T> = T & { computed?: undefined };
```

### LINKED ISSUE #361: [Proposal] - Supporting Generic properties in model
Sorry if this is a duplicate.

```
interface StoreModel<K extends {}> {
  data: K;
  act: Action<StoreModel<K>, void>;
}

function doSomething<K extends {}, S extends StoreModel<K>>(
  s: State<S>,
  a: Actions<S>
) {
  a.act();
  return s.data;
}
```

According to Typescript, `act` and `data` don't exist on `a` and `s`, respectively.

### LINKED ISSUE #386: Computed value from store state does not trigger update when using destructuring
Hi,

I came across this strange bug that I have isolated to be an issue with computed values if and only if the computed value is using the global store state AND you access the computed state property by object destructuring. In this case an update to the global store value does not trigger an update in components using the computed value.

I have created a small app that re-creates the problem here: https://gist.github.com/jakobmulvad/14226b9b6ea7a59644a181d8a43bc5e6

When you click the button you will see that the component that uses the computed value by destructuring are not updated when the action is triggered. If however the component is updated by any other means, it will also update to display the correct computed value.

Thanks for a great framework! I'm really digging it! ❤️ 

versions:
node: 12.13.1
react: 16.12.0
react-dom: 16.12.0
easy-peasy: 3.2.3

### LINKED ISSUE #390: create a middleware which executes before thunk middleware
Hi,

We are creating a middleware that stops unnecessary requests to the backend. But this is not stopping thunks from executing. My guess is thunk middleware is running before my custom middlewares. 

Here is the code

``` typescript
const loaderHash: { [key: string]: boolean } = {};

export const debouseAPIRequests: Middleware = store => next => (action: {
  type: string;
  payload: any;
}) => {
   if (action.type.indexOf("@thunk") === 0) {
    const key =
      action.type.replace(/(.+)(\((success|fail|start)\))/, "$1") +
      JSON.stringify(action.payload);


    if (action.type.endsWith("(start)")) {
      if (!loaderHash[key]) {
        loaderHash[key] = true;
        next(action);
      }
    }

    if (action.type.endsWith("(fail)") || action.type.endsWith("(success)")) {
      loaderHash[key] = false;
      next(action);
    }
  } else {
    next(action);
  }
};

```

### LINKED ISSUE #402: data state dose not want to be deep wrapped model
just want user is a data state not a modal

```js
interface User {
   name: string;
   getDisplayName:  () => string
}

interface UserModal {
  user: User;
  fetchUser: thunk(.....)
}
```

const user = useStoreState(state => state.userModal.user)

**the user is StateMapper\<User\> ???   just want the user is User not  StateMapper\<User\>**

### LINKED ISSUE #419: Add `effectOn` API
So I was trying to figure out a way to centralize an effect based on when a specific part of a model changes. This should cover when it loads from default to persisted value. The idea would be something like:

```tsx
const storeModel = persist({
  token: '',
  setToken: action((state, payload) => ({ ...state, token: payload })),
  onTokenChange: effectOn(state => state.token, async () => {
    // fetch user data and dispatch actions.
  })
})
```

This would allow us to fetch user data when token is changed from anywhere, whether it be from an action, or from the initial load from the persisted store.

The specific api isn't important, but I think this mechanism could allow much easier centralization of logic. 

What do ya'll think?

### LINKED ISSUE #427: Types error when creating a computed property based on another computed one
TypeScript fails to compile in a scenario similar to the following one, where a computed property is based on another one: 

```
test('computed properties can access global state', () => {
  // arrange
  const store = createStore({
    products: {
      items: [{ id: 1, name: 'boots', price: 20 }],
      itemMap: computed([state => state.items], items =>
        items.reduce((acc, cur) => ({ ...acc, [cur.id]: cur }), {}),
      ),
      setProductName: action((state, payload) => {
        const product = state.items.find(p => p.id === payload.id);
        product.name = payload.name;
      }),
    },
    basket: {
      productIds: [1],
      products: computed(
        [
          (state, storeState) => storeState.products.itemMap,
          state => state.productIds,
        ],
        (productMap, productIds) => productIds.map(id => productMap[id]),
      ),
    },
  });
```

In my case the first computed property resolves to the object Role, only if the id is provided, otherwise it's undefined.

The second computed property uses the `Role `as dependency and returns a string containing `Role.name`.

```
interface Role {
  id: number,
  name: string,
  description: string
}
```

```
interface State {
  currentRole?: Computed<State, Role, StoreState>,
  anotherComputed?: Computed<State, string, StoreState>
}

export const model: State = {

  currentRole: computed(
    [
      (_, storeState) => storeState.role.currentId,
      (_, storeState) => storeState.data.rolesMap
    ], 
    (roleId, rolesMap) => !!roleId ? rolesMap[roleId] : undefined
  ),
  currentRoleFullName: computed(
    [
      state => state.currentRole,
    ],
    (role) => role && role.name
  ),
}
```
First it complains that the first computed cannot be assigned to the type because it can be undefined (I marked the property as optional with an `?`, so normally it should be no problem).
`Type 'Computed<State, Role | undefined, StoreState>' is not assignable to type 'Computed<State, Role, StoreState>'.`

If I change the type of the first computed to `Computed<State, Role | undefined, StoreState>` then it complains at the second, even if I check if role is not undefined:
`Property 'name' does not exist on type 'Computed<State, Role | undefined, StoreState>'.`

How do I handle this case? Why does the type resolve to Computed<> and not the underlying result type?

### LINKED ISSUE #439: [Question] Accessing global store actions from component store thunk
I have this scenario where I have `SignInStore` that when successfully logged in, I need to save the current user globally in the context and what I currently did was created two exact copy of `setCurrentUser` in the component and global stores and used `useEffect` to handle communication between the two. Is there a way to access it directly from the component store?

### LINKED ISSUE #440: Support Map and Set with immer and/or immer-peasy
I just discovered that I can't use `Map` because [the changes](https://github.com/ctrlplusb/immer-peasy/compare/master...immerjs:master) are not included/rebased into `immer-peasy`.

Is that intentional? How can I help?

### LINKED ISSUE #444: AddModel Doesn't rehydrate Persist
Have a project I'm working on where we use a lot of addModel as we lazy load in different parts of our app. Went to persist some filtering options on one such Model but it never gets rehydrated.

I didn't see anything in the documentation about manually calling it to rehydrate for a specific model. Is there anything I missing?

Thanks.

### LINKED ISSUE #446: Computed properties fire when an action is emitted
I'm attempting to use computed properties to preserve fetch results and `createContextStore`;

```
...
getSomething1: computed((state) => fetch("something1", state.date)),
getSomething2: computed((state) => fetch("something2", state.date)),
getSomething3: computed((state) => fetch("something3", state.date)),
...
```

I'm then consuming only 1 of the computed properties; `getSomething1`, everything works perfectly until I change the state of 'date', then all three `getSomething` (1, 2, 3) fire off fetches even though I'm only using the first one anywhere.

### LINKED ISSUE #451: Introduce useLocalStore API, deprecate useComponentStore
Howdy all;

So, after some use and feedback I believe the the `useComponentStore` API is verbose and doesn't lend itself to more dynamic composition.

I am playing with a new API in mind and would like to get some feedback from the community before proceeding to build it out.

```typescript
import { useLocalStore, createStore, action } from 'easy-peasy';

export default function Counter() {
  const [state, actions] = useLocalStore(() => createStore({
    counter: 0,
    increment: action(_state => {
      _state.counter += 1;
    })
  }));

  return (
    <div>
      <h1>{state.counter</h1>
      <button onClick={() => actions.increment()}>Increment</button>
    </div>
  );
}
```

As you can see, you use the new API as a hook directly within your component. You need to provide a function that returns a new store instance.

This may appear slightly verbose at first appearance, however, using this approach allows me to support a secondary argument which would be a dependency array. Should one of the dependencies change, then the function would be executed again in order to create a new store instance.

```typescript
function EditProduct({ id, name }) {
  const [state, actions] = useLocalStore(
    () => createStore({
      id,
      name,
      editName: action((_state, payload) => {
        _state.name = payload;
      }),
    }),
    [id, name]
  );

  // ...
}
```

As you can see in this example, we have a component used to edit a products details. Every time the incoming props change we reinitialise the store, thereby representing the editing for a different product.

In addition to this, explicitly requiring that you use the `createStore` API allows you to provide a custom configuration to the underlying store.

Thoughts?

### LINKED ISSUE #452: Replace build system with microbundle
Everyone loves how small `preact` is. It's amazing what they fit into 3kb. I'd like us to move much more towards a smaller bundle size for v4. I investigated their strategies and it appears the `microbundle` packager/bundler can help with this quite significantly. Along with some babel magic. 😅

### LINKED ISSUE #454: How do I know when persist finishes saving changes?
On my logout flow, I want to clear the state so I have my model setup like:

```javascript
persist({
  session: {
    authenticationToken: null,
    login: action((state, payload) => {
     state.authenticationToken = payload.authenticationToken
    }),
    logout: action((state, payload) => {
      state.authenticationToken = null
    })
})
```

Then in my app I do something like:
```javascript

const App = () => {
  const logoutAction = useStoreActions(actions => actions.session.logout)

  const doLogout = () => {
     logoutAction()
     window.location = '/login'
  }

  return <button onClick={doLogout} />Logout</button>
}
```

I think I need some way to wait for the persist action to  complete before I do my redirect because I don't think it's  getting saved into sessionStorage (I still see authentication token in my store after I redirect).

Is there a good way to accomplish this?

### LINKED ISSUE #462: Computed properties execute even if they are not currently being accessed.
Hi @ctrlplusb,

We ran into an issue with computed properties. It seems like computed properties which are currently not being accessed also execute which cause all sorts of weird problems.

Here is a simple example:

````js

const userModel = {
  data: {},
  withBlocked: computed(
    [state => state.data],
    user => {
      // computed body
    },
  ),
  withFriends: computed(
    [state => state.data],
    user => {
      // computed body
    },
  ),
  authorize: action((state, payload) => {
    state.data = payload;
  }),
  unauthorize: action(state => (state.data = {})),
};
````

now somewhere else when we execute:

````js
const actions = store.getActions();
actions.user.unauthorize();
````

both `withBlocked` and `withFriends` will be fired even though they are not used anywhere.

This maybe related to https://github.com/ctrlplusb/easy-peasy/issues/446

I started to look at the source code and I think the issue could be related to: https://github.com/ctrlplusb/easy-peasy/blob/master/src/create-reducer.js#L48

Although this probably requires a bigger refactor in order to track which computed props are currently utilized.

### PR DESCRIPTION
**Pull request by ctrlplusb**
This PR is currently published as `easy-peasy@3.4.0-beta.2` 

We would appreciate early testing and feedback. 💜

The updated website for this PR can be found at https://easy-peasy-v3-4-0.now.sh/

### Deprecated `createComponentStore`

The API for `createComponentStore` was a bit verbose and limited. It will be removed in the next major release.

I plan to release non breaking changes for a while still, so no stress.

We have introduced a new API, `useLocalStore`, which replaces this one.

### Adds new `useLocalStore` hook

This API acts as a replacement of the deprecated `createComponentStore` API.

Allows you to create a store to represent the state for an individual React component. This is essentially an alternative to the official `useState` and `useReducer` hooks provided by React for component-level state.

```javascript
function MyCounter() {
    count: 0,
    increment: action(_state => {
      _state.count += 1;
    })
  }));

  return (
    <div>
      {state.count}
    </div>
  );
}
```

Please see the [new API documentation](https://easy-peasy-v3-4-0.now.sh/docs/api/use-local-store.html) for more information on this API.

Closes #451 
Closes #439 

### Adds new `unstable_effectOn` model API

> ***Note:** this is an experimental API. We are pre-releasing it to allow for early feedback. The API is subject to breaking changes with any release of Easy Peasy. As such we have prefixed the API with "unstable_", much like React does with its experimental APIs. Once the API has stabilised the "unstable_" prefix will be removed and semver based releases will be respected.*

Allows you to declare an effect within your model which will execute every time the targeted state changes.

Two arguments are provided to unstable_effectOn; namely the `stateResolvers` and the `handler`. The `stateResolvers` are an array of functions which should resolve the target state that should be tracked. When the tracked state changes the `handler` function is executed.

The `handler` can be *asynchronous* or *synchronous*. It receives the store actions, a `change` object containing the `prev` values for the targeted state and the `current` values as well as the `action` that caused the change in state. It additionally receives a `helper` argument allowing you to access the store state etc.

The `handler` additionally allows you to return a dispose function, which will be executed prior to the next execution of the `handler`. This can be useful in performing things like API call cancellation etc.

```javascript
import { unstable_effectOn } from 'easy-peasy';

const todosModel = {
  items: [],
  saving: false,
  setSaving: action((state, payload) => {
    state.saving = payload;
  }),
  unstable_effectOn(
    // Provide an array of "stateResolvers" to resolve the targeted state:
    [state => state.items],
    // Provide a handler which will execute every time the targeted state changes:
    async (actions, change) => {
      const [items] = change.current;
      actions.setSaving(true);
      await todosService.save(items);
      actions.setSaving(false);
    }
  )
};
```

Closes #419 

### TypeScript: Adds support for generics in models

Previously if you defined a model containing generic state, like below, TypeScript would break within your actions.

```typescript
interface StoreModel<K> {
  data: K;
  updateData: Action<StoreModel<K>, K>;
}

const numberStoreModel: StoreModel<number> = {
  data: 1337,
  updateData: action((state, payload) => {
    // A TypeScript would be thrown at this point
    //     👇
    state.data = payload;
  })
}
```

Unfortunately we were unable to directly resolve the case of generic properties due to current limitations with the TypeScript type system. We created a [StackOverflow question](https://stackoverflow.com/questions/58738700/issue-with-generic-properties-when-type-mapping) which details the problem.

In a gist; the issue is that Easy Peasy's underlying `State` and `Action` types map over the user provider model types in order to filter down to types that represent state and actions respectively. However, when defining a generic state, TypeScript assumes that the generic state intersects with types that are trying to be filtered out of each case. Therefore the filtering ends up always removing your generic state.

To resolve the case of generic state we have introduce a new API helper. Any time you wish to have a generic state value within your model, simply wrap it with the `Generic` type, and then assigned the associated value within the model instance using the `generic` helper.

```typescript
import { Generic, generic } from 'easy-peasy';

interface StoreModel<K> {
  data: Generic<K>;  // 👈
  updateData: Action<StoreModel<K>, K>;
}

const numberStoreModel: StoreModel<number> = {
  data: generic(1337),  // 👈
  updateData: action((state, payload) => {
    // Note that you don't need to wrap the payload with the
    // helper          👇
    state.data = payload;
  })
}

numberStoreModel.getState().data
// 1337

typeof numberStoreModel.getState().data
// number
```

Note how you only need to use the helper at the point of defining the initial value of the generic model state. Within your actions and anywhere you consume the state you would treat the value as the underlying generic value (i.e. a `number` in the example).

Closes #300
Closes #361 

### Improves the persist APIs on the store instances

We had undocumented APIs regarding persistence that were being exposed on the store instances. These are helpful in many cases, such as being able to flush persistence prior to navigating away from your application, or awaiting for persisted data to be rehydrated prior to rendering your application. We highly recommend you read the respective Store API docs.

Closes #454 

### Removes limit on TypeScript model mapping

Previously, if you had a model more than 6 levels deep, in terms of object structure, the TypeScript mapping wouldn't work as expected. This is no longer the case. 

I still think having that deep of a model is a bit excessive though. 😅

### Replaces `immer-peasy` with official `immer`

We have replaced our forked/patched version of immer with the official version. Thanks to their newly released support for computed properties. 🎉

Closes #462 
Closes #446 
Closes #440 

### Adds ability to await on rehydration of persisted data for dynamically added model

When utilising `persist` against a dynamically added model, i.e. via `store.addModel`, you may need to await on the rehydration due to utilising an asynchronous storage engine.

You can now do so via the returned `resolveRehydration` helper.

```typescript                             👇
const { resolveRehydration } = store.addModel('products', productsModel);
//            👆
// Deconstruct the returned object to get a handle on resolveRehydration

//     as we are using an asynchronous storage engine we will await the
// 👇 the promise returned by the resolveRehydration function.
resolveRehydration().then(() => {
  console.log('Rehydration is complete');
});
```

Closes #444 

### Moves internal `redux-thunk` binding to grant user defined middleware higher priority

This will allow you to influence thunks prior to their execution. For advanced cases. 😊

Closes #390

### TypeScript: Loads of fixes and improvements to the typings

The typings are being combed over multiple times and various fixes and improvements are being made. This is an ongoing task and you can expect many more improvements to be made still, including the addition of proper documentation on each type within the typings. This should improve the dev experience within your editor as you will get inline guidance on the APIs along with links to the official documentation for them.

### Fixes `persist` data not rehydrated for dynamically added models

Data persisted within models added via the `store.addModel` API were not having their data rehydrated. This is now fixed.

Closes #444

### Fixes `merge` and `mergeDeep` strategies for `persist` rehydration

A bug was identified where it was possible for persisted state to be misaligned with an evolving datamodel in terms of data types.

For example, you could have the following state persisted:

```json
{
  "counter": {
    "count": 1337
  },
  "todos": ["one", "two", "three"]
}
```

And since the state was persisted there occurred an update to the data model:

```typescript
const storeModel = {
  counter: {
    count: 1,
    increment: action(/* ... */)
  },
  todos: {
    items: ['one', 'two', 'three'], 
    addTodo: action(/* ... */)
  }
}
```

Note how the data structure for `todos` has evolved since the persisted state. It is now an object with nested properties.

Previously, for `merge` and `mergeDeep` it did not take into consideration the evolving store model and would rehydrate a persisted state despite it being misaligned in terms of tree data type structure. With the fix if a misaligned data type is found then the data specified on the store model is used instead of the persisted state.  

The logic for this ignores values where they are `null` or `undefined`, but otherwise will compare data types of the model vs persisted state to ensure the types match.  If the types do not match then the store model is used over the persisted value. So for our example above the rehydrated state would be the following:

```json
{
  "counter": {
    "count": 1337
  },
  "todos": {
    "items": ["one", "two", "three"]
  }
}
```

Note how `todos.items` matches the store model rather than the persisted state.

Moving forward we strongly recommend that the `mergeDeep` strategy is used when rehydrating state. This will ensure that the types are aligned right down the entire tree of the persisted state vs your store model.  

Closes #355 

### TypeScript: Fixes the statemapper eating up "classes"

If you assigned a class instance to your state the typings from `getState` / `useStoreState` etc would report your type as being something similar to:

Screenshot 2020-03-21 20 05 29

With this fix your state will correctly be reported back as the actual class type (`Person` in the examples case).

Closes #402

### TypeScript: Fixes action mapper so that state isn't display when viewing actions

The VSCode intellisense for actions was showing state. The types have been patched so that only actions will beb displayed.

### TypeScript: Fixes state mapper where actions were still being represented on state

There were cases if you had a nested action and only primitive state next to it that you would see the nested action. This is no longer the case.

### TypeScript: Fixes computed properties state resolvers not inferring resolved types correctly

If you used a state resolver array within your computed properties, whilst using TypeScript, the inferred types based on the resolved state was incorrect. This is now fixed.

Closes #427

### Website: Adds a known issue in regards to computed properties + Typescript 

Unfortunately, due to the way our typing system maps your model, you cannot declare a computed property as being optional via the `?` property postfix.

For example:

```typescript
interface StoreModel {
  products: Product[];
  totalPrice?: Computed<StoreModel, number>;
  //       👆
  // Note the optional definition
}

const storeModel: StoreModel = {
  products: [];
  // This will result in a TypeScript error 😢
  totalPrice: computed(
    state => state.products.length > 0
      ? calcPrice(state.products)
      : undefined
  )
}
```

Luckily there is a workaround; simply adjust the definition of your computed property to indicate that the result could be undefined.

```diff
  interface StoreModel {
    products: Product[];
-   totalPrice?: Computed<StoreModel, number>;
+   totalPrice: Computed<StoreModel, number | undefined>;
  }
```

### Fixes computed properties error for dynamically added model

Thanks goes to @jchamb for this fix. 💜

Computed properties were throwing errors intermittently when the `store.addModel` API was being used. This fix puts a guard in place to protect against the error.

### Fixes an error on the website -> docs -> quick-start.md

Thanks goes to @hualu00 for this fix. 💜

### Replaces `rollup` with `microbundle` to bundle library

Saves some more valuable bytes in bundlesize.

Closes #452 

### Updates website to include known issue on computed property destructuring

Closes #386 

### Improves the `persist` API documentation

Closes #454 

### Adds "Community Extensions" page to website

Closes #359


avevlad:
@ctrlplusb Looking forward to it, thanks!

onionhammer:
Boo. I liked createComponentStore.

ctrlplusb:
> Boo. I liked createComponentStore.

Did you try the replacement yet?

onionhammer:
@ctrlplusb not yet, I'll try it next week

ctrlplusb:
Ok, no worries. I am hoping that you will find it far more concise and useful. As it is described within your component you can hook into a context store or callback within your model now too. Far more powerful and composable.

onionhammer:
@ctrlplusb I used it and like it.

One thing that would be amaze is the ability to add a <Provider> to the output of useLocalStore, so nested child components could utilize the same store from a component.

GollyJer:
Where does...
```javascript
await todosService.save(items);
```
... come from in the `effectOn` example?

mkuklis:
@ctrlplusb thank you for your progress on this.  

I just wanted to give you a feedback. 

We upgraded to `3.4.0-beta.2` from `3.4.0-alpha.23` and unfortunately the issue with computed properties described in https://github.com/ctrlplusb/easy-peasy/issues/462 came back. 

Our setup is pretty straightforward with `"easy-peasy": "3.4.0-beta.2"` in package without direct immer dependency.

I do see immer `v7.0.8` installed.

I wonder if something has changed which is triggering this behavior.

ctrlplusb:
This will be replaced by v4. 😅

PR #461, opened 2020-03-26
Runtime Injection on Context Store -> V3.4

kamikazePT:
I simplified things from our last discussion, added support for this on 3.4

useLocalStore already supports this, so no changes there

useComponentStore is deprecated so I left it as is

- Enables dependency injection on context store's Provider at runtime
- Preserves previous state on store recreation based on injections being added/removed/edited

For example you have a "history" object from react-router and you have a thunk action that has some sort of guard that needs some sort of context about navigation, you inject "history" runtime in the Provider as a prop and your thunk actions can access it.

Or if you use react-i18next and you need the current language of i18n instance for some http request being made on a thunk, you can access it if you inject i18n instance as a "dependency" on the Provider

Or imagine that you have a AuthContextStore higher in the tree and you have a ProfileContextStore.
ProfileContextStore you can change the user name and submit it, changing the name of the "authenticated user". But since both stores are isolated, you can add a depedency to ProfileContextStore like:

<ProfileContextStore onProfileSubmit={AuthContextStore.useStoreActions(actions => actions.setUserProfile)

something like that... or a way to inject the current "access token" on thunk requests instead of having of having global instances ( for example using keycloakjs )

Not sure if I'm making myself clear about the idea, I just want a way to remove the need to export global instances for history or i18n or keycloakjs, I want them to be created along the way and provide them as dependencies for stores that might need them, instead of having imports of "global" instances at store level making them "hardcoded" to the project :/ when sometimes those context stores could be re-used in other projects

Edit : actually useLocalStore doesnt actually support preservativon of store state when dependencies change, or you dont agree with preservating the state?


jmalins:
Our team has great interest in this functionality, as we're currently (hackishly) mutating the `injections` object at runtime to achieve the same effect.

I have some concerns about the `...rest` spread of dependencies on the context provider:

 1. Clarity: it is not particularly clear / self-documenting that the spread props map to the `injections` object.
 2. Expandability: if the entire prop space of the context provider, other than `initialData`, is used for injections, then any added prop (config, for example) would be a breaking change due to possible name conflicts.
 3. Strong Typing: Typescript typing of mixin React component props is problematic at best.

In my opinion, an explicit `injections` prop on the context provider, which the end-user could memoize as necessary, would be clearer, have better forward compatibility and be easier to type.

Just my two cents.

kamikazePT:
On my previous PR regarding this feature I actually had injections prop and based on what you are saying I agree that I should have kept it, Will fix that :)

But passing in an injections prop, the user always should memoize it, otherwise it would recreate the store everytime. The options:

- internally check changes on injections props and not injections object itself
- add a new prop (func) like the nast PR called "shouldRecreateStore" with a default that does the point above and letting the user tune things up (might be interesting but I think it adds some unnecessary complexity)
- Just document that the user cant create the object directly on jsx and should memoize it since it Will be a different instance everytime (seems shitty user friendly)

What do you think?

jmalins:
In our codebase, we try to stick with single-responsibility principle... nothing memoized automatically. The team knows that if they need a shallow-equal memoized object literal, they need to make it themselves (using a one-liner hook). I agree with your last point though, that may not be the most novice-friendly approach for a publicly available library where a mistake erases your store.

For simplicity, I'd say go with option 1 and do a shallow-equal memo of the `injections` object internal to the provider. The `injections` object is broken apart prop-by-prop when mixed-in with the declare-time injections anyway, so prop-by-prop compare seems appropriate. This would also get rid of the `Object.values(...)` spread in the `useMemo` dependencies array (which probably always works, but is dependent on property enumeration order).

FWIW, this is our generic shallow-equal memo hook (pardon the Typescript), which could probably just be inlined:
```typescript
export const useShallowEqualMemo = <T>(value: T): T => {
  const valueRef = useRef(value);
  if (!shallowEqual(value, valueRef.current)) {
    valueRef.current = value;
  }
  return valueRef.current;
}
```

kamikazePT:
shadow equal comparison added and injections comes from an injections prop :)

I used your custom hook as well @jmalins 
I placed it on "lib.js" since it was the best suited file I found, I was going to put it on "hooks.js" but "hooks.js" only has specific store hooks, It seemed weird to put it there...

useLocalStore after looking better at it, It supports state persistence on dependency changes, so I guess that hook works out of the box for runtime injections :) No need to change anything for local store

@jmalins does this API solve your issue? Anything missing?
@ctrlplusb what do you think? I remember we also discussed having a "runtimeStoreConfig" where we could change some configurations, but besides injections and maybe name, I'm not so sure we need the added complexity for that subset of StoreConfig

ctrlplusb:
Sorry for the delayed feedback. I am slowly working my way towards this PR.

Lots going on for v3.4.0  right now. 😅

kamikazePT:
@ctrlplusb 
I just noticed that my latest merge commit to fix conflicts, the parenthesis on arrow functions were removed and staged as changes made by me (it ran eslint on pre-commit because it assumed as staged files) o.O did you update the eslint configuration and didn't run it aftwards for the repository?

kamikazePT:
@ctrlplusb 
Hey! been a while, going to pick this up soon and merge with your current developments, regarding the proposed API do you have any feedback you want to give?

ctrlplusb:
Hey all 👋

So I have integrated something like this into the next release, which is now v4.

Here are some related docs.

### Defining injections at runtime

It is possible to provide the store injections at runtime. In order to do so you
can provide them the store's `Provider`.

```javascript
import CounterStore from './stores/counter';

function MyApp({ language }) {
  // Imagine we had a value that changes at runtime, which we intend to use
  // as an injection into our application
  const translator = useTranslator(language);

  return (
    <>
      <Header />
      {/* pass down the injections into the provider
                                               👇  */}
      <CounterStore.Provider injections={{ translator }}>
        <Main />
      </CounterStore.Provider>
      <Footer />
    </>
  );
}
```

This will override any previous injections that were defined when creating the
store. If you wish to instead update the existing injections you can utilise the
function form.

```javascript
import CounterStore from './stores/counter';

function MyApp({ language }) {
  // Imagine we had a value that changes at runtime, which we intend to use
  // as an injection into our application
  const translator = useTranslator(language);

  return (
    <>
      <Header />
      <CounterStore.Provider
        {/* Pass down a function into the injections.
            It receives the previous injections.
                          👇 */}
        injections={(previousInjections) => ({
          // Spread the existing inject values so we can keep them:
          ...previousInjections,
          // Then overwrite the "translator":
          translator,
        })}
      >
        <Main />
      </CounterStore.Provider>
      <Footer />
    </>
  );
}
```

---

Thoughts on this approach?

ctrlplusb:
Closing this. Feeling ok with it, I'll detail this in the upcoming v4 PR.

kamikazePT:
@ctrlplusb looking forward to V4 :)

PR #463, opened 2020-03-29
Updated "Using typed injections" for better type safety

jankalfus:
Updated the "Using typed injections" section of the typescript tutorial to provide a more tightly bound example of typing the store injections. I believe this is more type safe than the original in the sense that there's now a single source of truth.


jmyrland:
Hi! I'm a new collaborator for easy peasy, and I'm currently going through open github issues & PRs.

I like this change! Are you able to resolve the conflicts?

jmyrland (approved):

PR #466, opened 2020-04-06
refactor(recipesDoc): renamed selector to computed

Matt-Swingvy:
I believe the name changed a while ago, but wasn't reflected in this doc.


ctrlplusb:
Very good catch, thanks. 💜

PR #469, opened 2020-04-08
shortened links

nainardev:
still works

but we should reduce number of links to the same app in same page :smile:


ctrlplusb:
Ah, thats a great trick. Good to know, thanks!

PR #514, opened 2020-07-08
Fix persist transformers order

czeslaaw:
Hi. While trying to make multiple transformers work (and almost losing my sanity) I encounter a bug with transformers order. Its caused by using transformers.reverse() method that mutates the array.

Expected result: transformers are applied **left to right** on persisting and **right to left** on rehydrating.

Actual result: transformers are applied **right to left** on persisting and **right to left** on rehydrating.

This fixes the issue and works according to https://easy-peasy.now.sh/docs/api/persist.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,9 @@
 <p align='center'>Vegetarian friendly state for React</p>
 <p>&nbsp;</p>
 
+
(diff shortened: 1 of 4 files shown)


ctrlplusb:
Oh wow this is a great catch! I am sorry it's taken so long to look into it. I'll address this asap.

ctrlplusb:
Amazing! I have applied this fix on v4.0.0, crediting you. 💜

Thanks so much for this one.

Issue #526, opened 2020-07-30
Initial data is undefined


onionhammer:
Provider restricts initialData to be `undefined` on the provider

![image](https://user-images.githubusercontent.com/969938/88950404-36699480-d25a-11ea-9d7c-d37960e4316c.png)

so now when I try to set initialData, i get a type error

![image](https://user-images.githubusercontent.com/969938/88950566-716bc800-d25a-11ea-919a-6fcc03e8446a.png)

If I force it using @ts-expect-error, it simply ignores the value and makes the initial data empty

onionhammer:
I think the types can be improved if 'createContextStore's default type for InitialData points to StoreModel instead of `undefined`

![image](https://user-images.githubusercontent.com/969938/88951223-4fbf1080-d25b-11ea-8ec5-94626514c1ad.png)

However, that still doesnt fix the fact that initial data doesnt actually work.

ctrlplusb:
Thanks for the report @onionhammer 

I'll look into this. 👍

ctrlplusb:
Sorted in v4 👍

szmarci:
@ctrlplusb I upgraded to v4 and I'm getting this error:
`Property 'initialData' does not exist on type 'IntrinsicAttributes & { runtimeModel?: PreFilterStore | undefined; injections?: PreFilterStore | ((previousInjections: PreFilterStore) => PreFilterStore) | undefined; }`
It has been removed?
![image](https://user-images.githubusercontent.com/1974290/97764493-1ecad200-1b0f-11eb-84e1-5fdf917a5c29.png)

onionhammer:
@szmarci This was a breaking change; I believe it was renamed to `runtimeModel`

szmarci:
@onionhammer Thanks, I look into it. I just didn't see it in the Release notes, under breaking changes section:
https://github.com/ctrlplusb/easy-peasy/releases/tag/v4.0.0

PR #527, opened 2020-07-30
Use `StoreModel` as the default type argument for createContextStore

onionhammer:
Use `StoreModel` as the default type argument for createContextStore's `InitialData` type

#526


ctrlplusb:
@onionhammer thanks for this. I am going to close this as it no longer applies in v4.0.0 😅

More info coming soon. 💜

PR #534, opened 2020-08-09
minor fix to docs: selector -> computed

kharandziuk:


ctrlplusb:
Thanks @kharandziuk this has already been fixed in v4 docs. 💜🙏

PR #540, opened 2020-08-22
don't persist computed fields, fixes #539

### LINKED ISSUE #539: computed properties ought not be persisted
When using `persist()` I'd expect computed properties not to be stored.

Root level computed properties are not persisted. But computed properties nested in the model are persisted as static values. I'd think computed values should never be persisted. 

In practice, the static property is overwritten with a computed getter in the rootReducer. So the statically stored value mostly just wastes space (and maybe causes a bit of confusion when debugging).

mighdoll:


ctrlplusb:
Awesome! Thanks for this @mighdoll 🙏

PR #554, opened 2020-09-28
Fix broken links

demiters:
I think this should do it.


ctrlplusb:
Thanks @demiters 

This has been fixed in the very soon to be released v4. 🙏

PR #558, opened 2020-10-05
Added missing possible word

w3bdesign:
Added missing possible word from the following sentence: 

"It is also to preload your store with some state by utilising the `initialState` configuration property of the store."


w3bdesign:
@ctrlplusb

ctrlplusb:
Thanks for this. I have fixed this in the latest docs which will be updated soon. 🙏

PR #572, opened 2020-10-16
Add fail to Helpers type

RobvH:
This PR updates the `index.d.ts` to add the `fail` function to the `Helpers` type.

**Motivation:**
I'm very excited about the way this new API works and am already leveraging it to pass payloads and errors to my error handling. 

**Problem:**
Calling `fail` from `helpers` required a `@ts-ignore` because the new `fail` function is not in the type declaration yet. 

**Solution:**
The interface of `fail` is set to `AnyFunction` because it receives any parameter and this ultimately is assigned to the `error` field of the thunk failure action payload.


.nvmrc
```diff
@@ -1 +1 @@
-8
+10
```
RobvH, line 1: Had to bump Node version to 10. I use a script that automatically changes my node version via nvmrc and `yarn build` fails with Node 8 because the currently locked version of `microbundle` uses promised based `readfile` calls, which are unsupported by Node 8.

ctrlplusb (approved):

ctrlplusb:
Thanks so much for this @RobvH 💜

PR #574, opened 2020-10-16
Document fail api use

RobvH:
This PR updates the `/docs/api/thunk` page to add the existence of the v4 `helpers.fail` function and document how to use it in thunk debugging. 

I had to code-dive to find out how to trigger failure status and was very happy to find an explicit function for this. It really let me use the thunk the way I wanted to, however, it would be fantastic to have it in the docs. 

I hope this helps!


ctrlplusb:
Top work. Thank you 💜

Issue #573, opened 2020-10-16
Issue with Immer polyfills after upgrading to version 4.0.1


jmyrland:
_First off: 👏 for the good work on the new release! I think I just had to change one line (in a test) after the upgrade for everything to be working 👏_

However, I've got a tiny issues with supporting older browser.  I've done as explained [here](https://easy-peasy.now.sh/docs/introduction/browser-support.html), to `enableES5()` for Immer - but it does not seem to be enough, as I still get an error (IE11):

![image](https://user-images.githubusercontent.com/837651/96282521-25b7f780-0fdb-11eb-8393-fd1404113b95.png)

The issue referred to, is [`20: "Cannot use proxies if Proxy, Proxy.revocable or Reflect are not available"`](https://github.com/immerjs/immer/blob/9257084c2b3c04fc9727e4c701d672483a7767b7/src/utils/errors.ts#L33).

Am I missing something, or is this an issue "upstream"?

ctrlplusb:
Hey @jmyrland 

That is music to my ears (except for the Immer part).

I downloaded an IE11 VM and did some immer deep diving. I believe I have fixed this now.

Could you please try:

```
npm install easy-peasy@4.1.0-alpha.7
```

It still needs the enableES5 👍

jmyrland:
Thanks for the quick patch! We are getting closer - the app now renders, but crashes due to `"[Immer] minified error nr: 19 ES5. Find the full error at: https://bit.ly/3cXEKWf"` ([`Immer error 19: "plugin not loaded"`](https://github.com/immerjs/immer/blob/9257084c2b3c04fc9727e4c701d672483a7767b7/src/utils/errors.ts#L32)):

![image](https://user-images.githubusercontent.com/837651/96417093-30e47080-11f1-11eb-8ba5-dcbe622b0305.png)

> 

I am loading & running `enableES5()` straight after running other react polyfills, followed by importing the rest and then rendrering the app:

```ts
import 'react-app-polyfill/ie11';
import 'react-app-polyfill/stable';

import { enableES5 } from 'immer';
enableES5(); // 👈

import { StoreProvider } from 'easy-peasy';
// ... import more stuff & render the app
```

Is this the correct sequence?

ctrlplusb:
Yes, that looks exactly right. I did the same in my test project and it was working. 🤔

Can you try install immer@7 as an explicit dependency?

ctrlplusb:
[...]

PR #575, opened 2020-10-19
Adds support for polyfilling required dependencies

jmyrland:
To polyfill (dependencies for) "easy-peasy":
```diff
-import { enableES5 } from 'immer';
-enableES5();
+import 'easy-peasy/polyfill/patch-immer';
```

```diff
-import { enableMapSet } from 'immer';
-enableMapSet();
+import 'easy-peasy/polyfill/patch-map-set';
```

To enable all polyfills:
```diff
+import 'easy-peasy/polyfill';
```

Benefits:
- Single entry for polyfill
- Can be extended over time if other polyfills are required
- Might also remove the need for installing `immer` explicitly, for this to work externally

Todo:
- [x] Update docs

related to #573


jmyrland:
@ctrlplusb I think this is ready for a test now, but I am not sure how to do this locally.

Is it much hassle for you to create a beta release of this PR, so I can test it? 👼

jmyrland:
I tried to just overwrite the folder in `node_modules` with the updated package, but it did not seem to work:

1. The `polyfill` folder is not included in the "dist", so for now it must be targeted via `easy-peasy/src/polyfill` - this is fixable, by relocating the files to the root.
2. It does not however solve the issue, as I get the following error:
    > Error: [Immer] minified error nr: 19 ES5. Find the full error at: https://bit.ly/3cXEKWf

Is this because of how easy-peasy is "built"? I.e. that immer is not "loaded" before easy-peasy is "loaded"?

The original idea of just "importing" the polyfill might not work, if it has to be "bundled" with easy-peasy, so it might be resolved in the same way Immer resolves it?

I'm not quite sure how to proceed with this 😔

jmyrland:
Closing this in favor of #576

PR #576, opened 2020-10-19
Adds direct polyfill support & updates docs

jmyrland:
Exposing the polyfill methods directly from easy-peasy,
enables users of this library to omit the "immer" dependency.

- Adds `polyfill.js`, which exports Immer polyfill methods.
- Rewrites section for `browser-support.md`.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,26 +1,26 @@
 {
   "index.js": {
-    "bundled": 46746,
-    "minified": 20221,
-    "gzipped": 6426,
+    "bundled": 56081,
+    "minified": 23293,
+    "gzipped": 6973,
     "treeshaked": {
       "rollup": {
-        "code": 209,
-        "import_statements": 168
+        "code": 139,
+        "import_statements": 135
       },
       "webpack": {
-        "code": 2607
+        "code": 2578
       }
     }
   },
   "index.cjs.js": {
-    "bundled": 58965,
-    "minified": 25523,
-    "gzipped": 7235
+    "bundled": 57629,
+    "minified": 24610,
+    "gzipped": 7162
   },
   "index.iife.js": {
-    "bundled": 61798,
-    "minified": 20702,
-    "gzipped": 6593
+    "bundled": 60584,
+    "minified": 20129,
+    "gzipped": 6533
   }
 }
--- a/index.d.ts
+++ b/index.d.ts
@@ -1117,3 +1117,11 @@ export function persist<Model extends object = {}>(
 export function useStoreRehydrated(): boolean;
 
 // #endregion
+
+
+// #region Polyfills
+
+export function enableES5(): void;
+export function enableMapSet(): void;
+
+// #endregion
--- a/src/index.js
+++ b/src/index.js
@@ -5,3 +5,4 @@ export * from './create-transform';
 export * from './provider';
 export * from './use-local-store';
 export * from './helpers';
+export * from './polyfills';
--- a/src/polyfills.js
+++ b/src/polyfills.js
@@ -1,0 +1,1 @@
+export { enableES5, enableMapSet } from 'immer';
--- a/website/docs/docs/introduction/browser-support.md
+++ b/website/docs/docs/introduction/browser-support.md
@@ -36,24 +36,20 @@ addTodo: action((state, payload) => {
 
 That's just a simple example. More complex/nested updates can get very tricky to manage.
 
-## Configuring Immer to support ES5
+## Supporting older browsers (IE11)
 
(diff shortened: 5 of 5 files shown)


jmyrland:
@ctrlplusb: I've refactored the implementation, and rebased the changes on top of `next`.

Ideally, I favor the solution in #575 - but I don't know how to implement this correctly with the bundler.

This also has a negative effect on the bundle size.  I'm not sure if these are "three-shake-able" though 🤔

It seems to be working now with "easy-peasy" as the only dependancy.

As you can see, I'm no expert in this field - so if you have some input, it's greatly appreciated 😅

ctrlplusb:
Hey @jmyrland 

Thanks for spending time to consider a useful solution to this problem. You were definitely on the right track.

I've gone ahead and done a dirty implementation of what you were asking for and have published it:

```
npm install easy-peasy@4.1.0-alpha.8
```

Would you mind trying something similar to:

```javascript
import "react-app-polyfill/ie11";
import "react-app-polyfill/stable";
import "easy-peasy/ie11";
import "easy-peasy/map-set";
```

Also, just a couple of tips in regards to this.

1. If you want to dev and test local packages then I highly recommend looking at [yalc](https://github.com/whitecolor/yalc). Doing local package dev is fraught with dangers given the design of the node module resolution algorithm. Symlinking is danger. Yalc gets it right by simulating a package repository experience.
2. You were close with your design. You just needed to add additional entry files instead of adding to `src/index.js`. So add a `ie11.js` at the root of the repo that will point to the build `dist/ie11.js` that came as a result of adding `src/entries/ie11.js` alongside an additional rollup config. Sorry, mouthful! 😀

jmyrland:
> Hey @jmyrland 
> 
> Thanks for spending time to consider a useful solution to this problem. You were definitely on the right track.
> 
> I've gone ahead and done a dirty implementation of what you were asking for and have published it:
> 
> ```
> npm install easy-peasy@4.1.0-alpha.8
> ```
> 
> Would you mind trying something similar to:
> 
> ```javascript
> import "react-app-polyfill/ie11";
> import "react-app-polyfill/stable";
> import "easy-peasy/ie11";
> import "easy-peasy/map-set";
> ```
> 
> Also, just a couple of tips in regards to this.
> 
> 1. If you want to dev and test local packages then I highly recommend looking at [yalc](https://github.com/whitecolor/yalc). Doing local package dev is fraught with dangers given the design of the node module resolution algorithm. Symlinking is danger. Yalc gets it right by simulating a package repository experience.
> 2. You were close with your design. You just needed to add additional entry files instead of adding to `src/index.js`. So add a `ie11.js` at the root of the repo that will point to the built `dist/ie11.js` that came as a result of adding `src/entries/ie11.js` alongside an additional rollup config. Sorry, mouthful! 😀

Sweet! I'll checkout the patch first thing tomorrow 👌

jmyrland:
I've tested `easy-peasy@4.1.0-alpha.8`, but I get an immer error when testing in IE 😟
> "[Immer] minified error nr: 19 ES5. Find the full error at: https://bit.ly/3cXEKWf"

ctrlplusb:
So this is f'ing weird to say the least. It looks like something strange is up with webpack and the module resolution. It actually completely ignores the import unless I export a function that gets executed outside the export.

Please try this version:

```
npm install easy-peasy@4.1.0-alpha.9
```

And this code:

```javascript
import "react-app-polyfill/ie11";
import "react-app-polyfill/stable";
import patchIE11 from "easy-peasy/ie11";
import patchMapSet from "easy-peasy/map-set";
patchIE11();
patchMapSet();
```

jmyrland:
> So this is f'ing weird to say the least. It looks like something strange is up with webpack and the module resolution. It actually completely ignores the import unless I export a function that gets executed outside the export.

Dang it!

> Please try this version:
> 
> ```
> npm install easy-peasy@4.1.0-alpha.9
> ```

Hooray 🎉 It's working now! The only thing missing is a type definition for the patch methods:

image

I solved it by `@ts-ignoring` the import for now.

ctrlplusb:
Ah, I just had an epiphany on why this might be the case... will try something out soon.

ctrlplusb:
Ok, one final try...

```
npm install easy-peasy@4.1.0-alpha.11
```

With just this format:

```javascript
import "react-app-polyfill/ie11";
import "react-app-polyfill/stable";
import "easy-peasy/ie11";
import "easy-peasy/map-set";
```

jmyrland:
Hooray 🎉 It's working! Nice job @ctrlplusb  👏

ctrlplusb:
Thanks for helping me drive this forward. Much appreciated! 💜

PR #592, opened 2020-10-30
Fix types fail helper

Cytren:
Added fail type definitions to `Helpers` and `thunk`.


ctrlplusb:
Thanks, this is fixed in the latest types. 🙏

PR #598, opened 2020-11-10
Expose storeState in effectOn dependency handlers.

### LINKED ISSUE #596: storeState argument on effectOn's stateResolvers is missing
According to docs, stateResolvers pass two arguments, so one can do something like this:
```
  const store = createStore({
    listener: {
      onTodosChanged: unstable_effectOn([(_state, storeState) => storeState.todos], (actions) => {
        actions.setFired(true);
      }),
      fired: false,
      setFired: action((state, payload) => {
        state.fired = payload;
      }),
    },
    todos: [],
    addTodo: action((state, payload) => {
      state.todos.push(payload);
    }),
  });
```

Unfortunately, the following test fails:
```
  expect(store.getState().listener.fired).toBe(false);
  store.getActions().addTodo('add onEffect api');
  expect(store.getState().listener.fired).toBe(true); // actual :false
```

methyl:
Also expose getActions in helpers.

Fixes #596.


methyl:
@ctrlplusb can you take a look?

surethink:
> @ctrlplusb can you take a look?

can this be done?

rbrtbrnschn:
@ctrlplusb , Please take a look at this 👍🏻 
closed [my issue](https://github.com/ctrlplusb/easy-peasy/issues/738) in favor of this pr.

rainerbruemmer:
@ctrlplusb Can you review this PR, please?

ctrlplusb:
Nice. I'll deal with the conflict for ya this week. Let's get this lined up for merge. :)

src/effects.js
```diff
@@ -50,6 +50,7 @@ export function createEffectHandler(
     const helpers = {
       dispatch: references.dispatch,
       getState: () => get(definition.meta.parent, references.getState()),
+      getActions: () => get(definition.meta.parent, _actionCreators),
```
no-stack-dub-sack, line 53: @methyl Is `getActions` possibly redundant / what is the use-case for using `getActions` over `actions`?  `getStoreActions` makes sense because otherwise we have no way to access the top-level store's actions, however wouldn't `getActions()` and the first argument of the handler function (`actions`) be identical here? 

With this added, I think there would be three ways to access the local store's actions: 1) via the `actions` argument, 2) `getStoreActions().localStoreName`, and 3) `getActions()`.

I believe that adding this as a helper also makes it inconsistent with `thunk` and `thunkOn` which only expose `getStoreActions`.

Just wanted to point this out and see if I may be missing something here. Thanks for adding this PR, though! I also noticed that `storeState` was `undefined` when playing around with this api. 

cc: @ctrlplusb

jmyrland: I agree. Given that the store actions are provided via the first argument for `effectOn`, I do not see the need for a `getActions` helper.

As you said; `getStoreActions` is really handy to expose 👍

methyl: Hey, not sure as it was quite long time ago, but maybe it's because `actions` can get stale somewhere in between creating effect handler and the handler being called? And `getActions` will always return fresh actionCreators. If you don't think it's possible, let's get rid of it to keep the API simple.

no-stack-dub-sack: Hey @methyl, @ctrlplusb asked me to clean up the conflicts on this branch and based on the discussion with @jmyrland I removed the `getActions` helper as well. Looks like this PR is ready for merging if you're good with this.

no-stack-dub-sack:
@ctrlplusb @jmyrland resolved the conflicts here. fwiw the VS Code diff is much clearer:

image

jmyrland (approved):
LGTM

src/effects.js
jmyrland: As long as the test passes (based on the example from #596) - we should be good :+1:

methyl:
LGTM too, thanks for your work!

jmyrland:
Thanks for your contribution @methyl 👏 This will be included in the release this week!

PR #722, opened 2022-01-30
Replace link on line 10

Aniganesh:
The link for the word "localStorage" used to lead to mdn page on Session storage instead of the mdn page on local storage. Replaced the link


ctrlplusb:
I appreciate the fix. ♥️

PR #735, opened 2022-02-23
Update computed property docs

no-stack-dub-sack:
The docs still reflect that computed props cannot be accessed from within actions, however this was addressed [here](https://github.com/ctrlplusb/easy-peasy/commit/772f1b5dffc1d98aff0339ffc5c750154886bc04), in the v 5.0.2 release.

Update docs to reflect that while computed props can be accessed from within actions, they will return the last cached value, rather than be recomputed.

### DIFF
--- a/website/docs/docs/api/computed.md
+++ b/website/docs/docs/api/computed.md
@@ -179,7 +179,7 @@ issues.
 
 ## Limitations and Known Issues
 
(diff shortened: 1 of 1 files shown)


w3bdesign (approved):

ctrlplusb:
♥️

PR #749, opened 2022-05-01
Typo and Added commas

yethranayeh:
- A small typo fix in the Action arguments section.
- Added some commas for better legibility.


ctrlplusb:
♥️

Issue #732, opened 2022-02-19
Reference to a computed property seems to trigger unnecessary re-renders 


ROTGP:
Using `v5.0.4`, I've noticed that making a reference to computed store properties causes the component making the reference to re-render, even when unrelated properties are being updated. I've made a demo to illustrate: 

https://codesandbox.io/s/interesting-morning-0qbd4u

Open the console and notice the following output (which is correct):

```
rendering FOO
rendering BAR
rendering BAZ
```

Now click on either of the two `Increment counter` buttons (one in the `Bar` component, and the other in the `Baz` component), which will increment the store's `counter` property. Note that the `Foo` component is the only component to make reference to `counter`, and as such you'd only expect `rendering FOO` to be output. However, you'll notice the following in the console output: 

```
rendering FOO
rendering BAR
```

Commenting out the reference to `completedTodos` (a computed property) in the `Bar` component will result in it not being mistakenly re-rendered when `counter` is updated. 

`Baz` component has a reference to `todos` (a standard, non-computed property) which does not result in a re-render when `counter` is updated.

In summary, having a reference to a computed property is causing the referring component to re-render, when I believe it shouldn't. Making a similar reference to a non-computed property does NOT cause the re-render (even when the computed property refers only to the non-computed property).  Is this expected behaviour? Is there something I'm missing?

no-stack-dub-sack:
@ROTGP, yes, I agree something seems off here. I recently started carefully analyzing some components for performance and noticed the same issue and was coming to report it when I saw you beat me to it.

@ctrlplusb to expand on the example provided by the OP, also see this sandbox: https://codesandbox.io/s/serene-leavitt-fg9p5l

I've added an additional computed property and a hook which tracks and logs when a component's deps change. According to [the docs](https://easy-peasy.dev/docs/api/computed.html#arguments) it should not be necessary to use state resolvers unless we want to access state from a different slice of our model: 

> In general it is recommended that you only use state resolvers if you need to resolve state from another part of your model. There are some performance benefits to be had by isolating local state, but in almost every case this would be insignificant.

[...]

PR #764, opened 2022-09-14
Optimize computed props

jmyrland:
An attempt to optimize computed props, by improving comparison of inputs & by checking if the value needs an update.

The value comparison uses a naive comparison of of the JSON stringified values.

This will support more complex inputs & avoid unneccessary updates for complex results (arrays, objects).

related to #732

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,27 +1,27 @@
 {
   "index.js": {
-    "bundled": 51221,
-    "minified": 21486,
-    "gzipped": 6458,
+    "bundled": 51519,
+    "minified": 21541,
+    "gzipped": 6505,
     "treeshaked": {
       "rollup": {
-        "code": 135,
-        "import_statements": 131
+        "code": 163,
+        "import_statements": 159
       },
       "webpack": {
-        "code": 2574
+        "code": 2635
       }
     }
   },
   "index.cjs.js": {
-    "bundled": 52597,
-    "minified": 22671,
-    "gzipped": 6630
+    "bundled": 52982,
+    "minified": 22785,
+    "gzipped": 6689
   },
   "index.iife.js": {
-    "bundled": 55510,
-    "minified": 18414,
-    "gzipped": 6030
+    "bundled": 55887,
+    "minified": 18499,
+    "gzipped": 6074
   },
   "ie11.js": {
     "bundled": 102,
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.17.2",
+    "fast-deep-equal": "^3.1.3",
     "immer": "^9.0.12",
     "redux": "^4.1.2",
     "redux-thunk": "^2.4.1",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -72,6 +72,7 @@ function createIIFEConfig(input, output, globalName) {
         redux: 'Redux',
         'redux-thunk': 'reduxThunk',
         immer: 'immer',
+        equal: 'fast-deep-equal/es6',
       },
     },
     external,
--- a/src/computed-properties.js
+++ b/src/computed-properties.js
@@ -1,9 +1,11 @@
 import equal from 'fast-deep-equal/es6';
+import { areInputsEqual } from './lib';
 
 export function createComputedPropertyBinder(parentPath, key, def, _r) {
   let runOnce = false;
   let prevInputs = [];
   let prevValue;
+  let prevStoreState;
 
   let performingEqualityCheck = false;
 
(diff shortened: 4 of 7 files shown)


jmyrland:
What do you think about this @ctrlplusb ? I haven't testet this against our codebase, but it should work in theory - and should avoid some unnecessary re-renders for computed props.

ctrlplusb:
Hey @jmyrland!

Thanks for jumping in with this PR. Interesting approach. I'd love to test this out in an alpha and see how it performs against some real world projects.

I'll try have a little play. :)

ctrlplusb:
Sure thing.

So in your investigation we _had_ to perform a deep equality comparison?

I've used this library at a previous client which is pretty blazing;

https://www.npmjs.com/package/fast-deep-equal

ctrlplusb:
Aha, I think I just made a good discovery.

When doing comparison the computed property being checked is included in the comparison. This should not be the case. It causes stack overflow issues and could be the root cause for some of the strange edge cases we are seeing around computed properties.

ctrlplusb:
@jmyrland 

Can you please test out `easy-peasy@5.1.0-alpha.1`

👍

jmyrland:
> Sure thing.
> 
> So in your investigation we _had_ to perform a deep equality comparison?
> 
> I've used this library at a previous client which is pretty blazing;
> 
> https://www.npmjs.com/package/fast-deep-equal

My theory is that computed arrays & objects are always updated, even if the value is the same - because of the reference comparison (objects & arrays will always fail due to the `===` comparison).

My attempt to resolve this, was to compare the new & previous computed result before updating the value. In my head, this should avoid unnecessary re-renders for computed arrays & objects.

I'll try out the alpha for the sandbox & report back 👍

jmyrland:
Dang it - it doesn't work for the [sandbox](https://codesandbox.io/s/long-cdn-mbbf9x?file=/src/components/Bar.tsx).

I was expecting "Bar render" to be 0.

Question: what default state resolvers are used (if not passing in any) to the computed prop?

jmyrland:
It is working - I was just referencing the incorrect alpha version 😁 

image

Nice one 🎉🎉

jmyrland:
We're using a proxied package manager, and I'm not able to find the updated alpha-version. Unable to test it "for real" atm.

jmyrland:
> Awesome!!
> 
> This confirms my thesis. The solution can actually be further optimised. 

Sweet!

>That being said I am gonna merge and release this as a patch so we can track associated bugs around computed properties. If this strategy resolves them I will do the optimisation refactoring.
> 
> Will do this in an hour or two. 👍

I was able to update to the alpha now for our apps - but I sadly get alot of failing tests (~50%) 😢 

It seems that computed props are `undefined` in most cases when a value is expected (it worked previously).

I am unable to test the actual app - as I end up in "login"-redirect loop, as we probably are using some computed props in the auth store.

Maybe [this is the culprit](https://github.com/ctrlplusb/easy-peasy/pull/764/files#diff-8a93e45e7019b6bd3b200c3f96dd05ee9c723b97eff6d09490ef51f5a7bb752eR22-R24)?

jmyrland:
Actually, using `areInputsEqual` just for checking inputs and `areEqual` for checking if the result changed should be sufficient for avoiding rerenders.

I tested this, and it seems to be working. I.e. if this approach is used, it is not a breaking change - and consumers do not have to update their state resolvers.

ctrlplusb:
Thanks for your help confirming all this!

Can you try `easy-peasy@5.1.0-alpha.2`?

w3bdesign (approved):

jmyrland:
> Thanks for your help confirming all this!
> 
> Can you try `easy-peasy@5.1.0-alpha.2`?
> 
> @jmyrland

A total of 1400~ passing unit tests, without changing a single line 👌🎉
And the [sandbox](https://codesandbox.io/s/naughty-khorana-kevtcf?file=/package.json) still works as expected.

You are a wizard @ctrlplusb ! 😁

ctrlplusb:
Haha, thanks for the approval @w3bdesign - were you having issues too? Did you try out the alpha release? If so did it resolve your issues?

ctrlplusb:
> 

Woooooooot!

Thanks so much for having such an extensive test suite for us to leverage!

jmyrland:
Happy to help! Also, this may close #732

no-stack-dub-sack:
@jmyrland @ctrlplusb Wow! Thanks for the work on this! This is going to greatly simplify a lot of code and reduce the need for state resolvers when trying to avoid un-wanted re-renders. And yes @jmyrland, from what I can tell, the sandbox is working perfectly with `completedTodos` and `safeCompletedTodos` both behaving as expected now. 

This is great!

PR #495, opened 2020-05-25
feat. Pass debug trace options to dev tools

el-gringo:

PR #771, opened 2022-09-16
feat: Pass debug trace options to dev tools

### LINKED ISSUE #678: Should configure redux dev tools to serialize map/set when map/set functionality is enabled.
Seeing as we are telling easy-peasy that we want to enable map/set functionality, it would make sense that redux dev tools should be configured to display them out of the box. I was surprised when most of my state wasn't appearing in dev tools.

A possible plan would be:

1. `devTools` store config can take a redux dev tools config object.
2. Pass that config when initializing redux dev tools https://github.com/ctrlplusb/easy-peasy/blob/5c4c0e8e0a708e0c218f82d0a89aa679d1f44dcb/src/create-store.js#L77
3. If no config object was set, and map/set is enabled, include something like this:
```js
  serialize: { 
    'map': true,
    'set': true,
  }
```

jmyrland:
continuation of #495 rebased on master

closes #678

### DIFF
--- a/examples/reduxtagram/src/store.ts
+++ b/examples/reduxtagram/src/store.ts
@@ -4,7 +4,7 @@ import posts from '@/data/posts';
 import { storeModel } from '@/model';
 
 const store = createStore(storeModel, {
-  devTools: import.meta.env.NODE_ENV === 'development',
+  devTools: import.meta.env.DEV ? { name: 'This is a test' } : false,
   initialState: { commentsModel: { comments }, postsModel: { posts } },
 });
 
--- a/examples/reduxtagram/src/store.ts
+++ b/examples/reduxtagram/src/store.ts
@@ -4,7 +4,7 @@ import posts from '@/data/posts';
 import { storeModel } from '@/model';
 
 const store = createStore(storeModel, {
-  devTools: import.meta.env.DEV ? { name: 'This is a test' } : false,
+  devTools: import.meta.env.DEV ? { name: 'Easy Peasy Reduxtagram' } : false,
   initialState: { commentsModel: { comments }, postsModel: { posts } },
 });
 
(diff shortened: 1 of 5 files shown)


w3bdesign (approved):

w3bdesign (dismissed):

src/create-store.js
```diff
@@ -21,7 +21,7 @@ export function createStore(model, options = {}) {
   const modelClone = clone(model);
   const {
     compose,
-    devTools = process.env.NODE_ENV !== 'production',
```
jmyrland, line 24: Any type updates we need to consider here?
Should we support both bool and object input?

src/create-store.js
```diff
@@ -76,6 +76,7 @@ export function createStore(model, options = {}) {
     window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__
       ? window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__({
           name: storeName,
+          ...devTools,
         })
```
jmyrland, line 80: Is this enough?

src/create-store.js
ctrlplusb: Hmmm. Good point. Will have to review.

jmyrland: I think I've fixed it now - but I was unable to verify it 😕 

Tried to install `easy-peasy` by using `file:../../` as the version for the example - but I just got the `5.1.0` source. And yes, I did run `yarn build` before doing this.

examples/reduxtagram/src/store.ts
```diff
@@ -4,7 +4,7 @@ import posts from '@/data/posts';
 import { storeModel } from '@/model';
 
 const store = createStore(storeModel, {
-  devTools: import.meta.env.NODE_ENV === 'development',
+  devTools: import.meta.env.DEV ? { name: 'This is a test' } : false,
```
ctrlplusb, line 7: **question:** Is this ternary intentional?

ctrlplusb:
Amazing, just a minor question. 👍

ctrlplusb (approved):
🚀

Issue #599, opened 2020-11-10
Using custom storage engine on React Native


ghost:
For some reason I cannot seem to be able to make my custom storage engine working on RN. What works in a clean CRA web project doesn't in a clean CRNA project.

I have created a new project by running `expo init` and choosing the Bare workflow with minimal settings. I've installed the `easy-peasy` package (4.0.1) and modified the App.js as such:

```
import React from "react";
import { Button, StyleSheet, View } from "react-native";
import {
  action,
  createStore,
  persist,
  StoreProvider,
  useStoreActions
} from "easy-peasy";

const store = createStore(
  persist(
    {
      count: 1,
      add: action(state => {
        state.count += 1;
        console.log("STATE CHANGE", state.count);
      })
    },
    {
      storage: {
        getItem: key => {
          console.log("GET_ITEM", key);
        },
        setItem: (key, value) => {
          console.log("SET_ITEM", key, value);
        },
        removeItem: key => {
          console.log("REMOVE_ITEM", key);
        }
      },
      allow: ["count"]
    }
  )
);

const MyButton = () => {
  const add = useStoreActions(actions => actions.add);
  return <Button title="Press" onPress={() => add()} />;
};

export default function App() {
  return (
    <StoreProvider store={store}>
      <View style={styles.container}>
        <MyButton title="Press" />
      </View>
    </StoreProvider>
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    alignItems: "center",
    justifyContent: "center"
  }
});
```

As I launch the project in my iOS simulator, `GET_ITEM [EasyPeasyStore][0]` is logged.
As I press the button, `STATE CHANGE 2` is logged.
However, `SET_ITEM` never gets logged.

As mentioned, the same code on a web project works fine and `SET_ITEM...` is logged.

Any ideas?

yixuan98:
Same issue after upgrading to v4.0.1, `setItem()` of custom storage isn't called when changing persisted state on RN 0.63.3.

luziczek:
I can confirm the same and also AsyncStorage doesn't seem to work. 
On v4.1.0-beta.4 also not working with RN 0.63.3.

Thram:
Same here with Easy Peasy 4.0.1

GollyJer:
Same. Just tried to switch over to easy-peasy's `persist`. It seems so much easier but kept running into issues before finding this thread. Should have searched first!

GollyJer:
[...]

PR #776, opened 2022-09-21
docs: Update reactotron recipe

### LINKED ISSUE #757: How to add reactron
https://github.com/infinitered/reactotron/blob/master/docs/quick-start-react-native.md

jmyrland:
Based on the existing docs & [this SO post](https://stackoverflow.com/questions/72947271/how-to-use-reactotron-redux-with-easy-peasy).

closes #757

### DIFF
--- a/website/docs/docs/recipes/connecting-to-reactotron.md
+++ b/website/docs/docs/recipes/connecting-to-reactotron.md
@@ -1,16 +1,18 @@
 # Connecting to Reactotron
 
-[Reactotron](https://github.com/infinitered/reactotron) is a desktop app for inspecting your React JS and React Native projects.
+[Reactotron](https://github.com/infinitered/reactotron) is a desktop app for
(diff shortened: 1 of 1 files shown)


jmyrland:
I've excluded the `window.requestIdleCallback = null;` which might be essential for it to work on iOS.

1. Should we include it in the recipe?
2. Should we try to find a solution for [this issue](https://github.com/ctrlplusb/easy-peasy/issues/599#issuecomment-781258630)

ctrlplusb:
Awesome. Perhaps we can just add a comment stating the potential issues around iOS with a link to the corresponding issue/discussion?

jmyrland:
> Awesome. Perhaps we can just add a comment stating the potential issues around iOS with a link to the corresponding issue/discussion?

Added the "solution" as part of the snippet, with a link back to the issue

ctrlplusb (approved):
LGTM :)

PR #773, opened 2022-09-17
chore: Extend examples

jmyrland:
- [x] Kanban (with tests)
- [x] Next.js

closes #699

PR #780, opened 2022-09-23
Add Reduxtagram example

### LINKED ISSUE #775: Add "reduxtagram" as an example
Port the [reduxtagram-easy-peasy](https://github.com/crissdev/reduxtagram-easy-peasy) into [examples](https://github.com/ctrlplusb/easy-peasy/tree/master/examples) using the latest version.

[Report back](https://github.com/crissdev/reduxtagram-easy-peasy/issues/3) when this is completed.

crissdev:
Rewrite [reduxtagram-easy-peasy](https://github.com/crissdev/reduxtagram-easy-peasy) and move it into Easy Peasy example directory.
The example is implemented with:
* Easy Peasy (of course!)
* React
* TypeScript
* Vite
* Vitest

@jmyrland It seems the upgrade to latest version was pretty straightforward. Let me know if there's anything needed for this PR. 
I think it would be a good idea to deploy the examples so that people can see how the app looks without having to clone the repo locally.

Closes #775


crissdev:
It seems the `test` script tries to run tests from the example directory. 😬

jmyrland:
> It seems the `test` script tries to run tests from the example directory. 😬

True - this is ignored in #

jmyrland:
> @jmyrland It seems the upgrade to latest version was pretty straightforward.

Awesome 🎉 Thanks for the contribution.

> I think it would be a good idea to deploy the examples so that people can see how the app looks without having to clone the repo locally.

Agreed - I've used [codesandbox](https://codesandbox.io/) for the other examples. Is there anything in this example that prevents this?

jmyrland:
> It seems the `test` script tries to run tests from the example directory. 😬

Ah, this is fixed in #773 . [`examples` must be included for the `testPathIgnorePatterns` property in package.json.](https://github.com/ctrlplusb/easy-peasy/pull/773/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R151)

crissdev:
> Agreed - I've used [codesandbox](https://codesandbox.io/) for the other examples. Is there anything in this example that prevents this?

Nothing prevents it but is there a way to tell codesandbox to use a specific directory? Otherwise having a code duplication is less than ideal because it requires to remember to update the codesandbox example whenever the code in the repository is touched - making maintenance a burden.

jmyrland:
> is there a way to tell codesandbox to use a specific directory? 

AFAIK this is not a feature in codesandbox.

> Otherwise having a code duplication is less than ideal because it requires to remember to update the codesandbox example whenever the code in the repository is touched - making maintenance a burden.

Yes, I completely agree - but I haven't found a better way to handle this. Examples are however relatively static, and it is better to have a live "sandbox" than nothing at all.

If you have a better way of solving this, I'm all ears 😄

crissdev:
@jmyrland I tried several times to create the CodeSandbox for this example but didn't manage for have the app working. I even tried to fork one of your examples and still failed 😬. So I looked up for available options.

Looking at Vercel's docs it seems they have [support for monorepos](https://vercel.com/docs/concepts/monorepos) but they enforce a [limit of 3 projects for each repository](https://vercel.com/docs/concepts/limits/overview).

For example, the link below is configured to deploy the Reduxtagram example from the `docs/reduxtagram-example` branch from `crissdev/easy-peasy` repository.

http://easy-peasy-example-reduxtagram.vercel.app/

----
So, I guess it's up to you to decide who the winner is this case:
1. CodeSandbox seems a good option because it's free and maybe the examples change very infrequently.
2. Vercel's monorepo support is easy to configure, deploys automatically whenever a comit is pushed, but has a limit of 3 projects which may not be enough.

jmyrland:
> @jmyrland I tried several times to create the CodeSandbox for this example but didn't manage to have the app working.

Yeah I had some issues the first time as well. I think it might be to a cold start. It worked after a while.

> So, I guess it's up to you to decide who the winner is this case:
> 1. CodeSandbox seems a good option because it's free and maybe the examples change very infrequently.
> 2. Vercel's monorepo support is easy to configure, deploys automatically whenever a comit is pushed, but has a limit of 3 projects which may not be enough.

I'd prefer sandboxes as you get to see both the source and the result - which is ideal for this type of example.

For displaying the result, vercel is a good option - but it won't scale with multiple examples as you pointed out.

I'll play around with your example tomorrow and do a review, so we can get this merged.

jmyrland (approved):
Did some minor, non invasive changes.

Also added a link to a [sandbox](https://codesandbox.io/s/ztuxzk).

Thanks for taking the time to add this 👏

ctrlplusb:
This amazing! Thanks so much for this contribution @crissdev ♥️

crissdev:
It's great to see this merged 🎊

Issue #628, opened 2021-02-19
Ability to use reducer-style action with Immer


allpro:
I have a large dataset that requires maximum update performance. This object is never mutated; only _replaced_ with a new dataset. A normal `action()` takes _many seconds_ to update state, even though it's a simple value setter...
````js
action((state, data) => {
  state.data = data;
})
````
It's Immer that is slow so I tried a reducer-style action to _bypass_ Immer. Since `data` is always a new object, immutability is automatic anyway. 
````js
action((state, data) => ({ ...state, data })
````

However I quickly found the return-state style does not work when Immer is enabled and there is a computed value (getter) in the model. This limitation seems to be in both v3 and v4. (I'm still using v3)

**My request is some way to _bypass_ Immer in an action.**
I _could_ use a `reducer()` method for this value and then `dispatch()` an action in the thunk to set it, but this syntax seems clumsy and non-standard. (But is what I'm doing for now.)

One obvious option is for the standard reducer-style return-state to work _without_ disabling Immer. 

Alternatively a _partial state_ could be returned from an action, which would be shallow-merged into state. This avoids having computed values included in the returned state data. For example...
````js
const model = {
  data: {},
  other: [],
  item: computed((state, id) => state.data[id]),
  setData: action((state, payload) => ({ data: payload }),
  fetch: thunk((actions, payload) => {
    getSomething(payload)
      .then(actions.setData);
  })
````
This could work like:
- The `setData()` action returns `{ data: {...} }`
- The model's `state` object is cloned for immutability
- The returned partial state is shallow-merged into state

This achieves the equivalent of `{ ...state, data: {...} }`, but without the issue of computed values inside the state object. Just one idea to work around the current problem.

**Comment by avevlad** [comment]
Same issue, why immer or easy-peasy so slow set state?

Store:

```tsx
// analogs.store.ts
setAnalogs: action((state, payload) => {
    state.analogs = payload;
}),
```

Handle click button action in react component:

```tsx

console.time("dgb: http request");
const response = await client.post(
  ENDPOINTS.assessmentApi + `/3564/analogs`,
  payload
);
console.timeEnd("dgb: http request");

console.time("dgb: setAnalogs");
// setAnalogsAsReactUseState(response.data.data.analogs);
assessmentActions.setAnalogs(response.data.data.analogs);
console.timeEnd("dgb: setAnalogs");

```

#### console.time setAnalogs `1260ms` :(

[...]

PR #781, opened 2022-10-02
Feat/action config (ability to pass a config object and disable immer) #628

s900mhz:
Hey all, 

This PR is to address this issue and milestone right here #628

Test were added, it's late here.. I didn't write tests for custom reducers. I can update this PR later with that.

### DIFF
--- a/website/docs/docs/api/action.md
+++ b/website/docs/docs/api/action.md
@@ -48,6 +48,13 @@ An `action` is a function that is described below.
 
     The payload, if any, that was provided to the
     [action](/docs/api/action.html) when it was dispatched.
+  
+  - `config` (Object)
+
+    The config object that was provided to the
+    [action](/docs/api/action.html) when it was dispatched. 
+    - `immer` (Boolean)
+      Whether to use `immer` to update the state. Defaults to `true`.
 
 ## Tutorial
 
--- a/website/docs/docs/api/action.md
+++ b/website/docs/docs/api/action.md
@@ -55,6 +55,8 @@ An `action` is a function that is described below.
     [action](/docs/api/action.html) when it was dispatched. 
     - `immer` (Boolean)
       Whether to use `immer` to update the state. Defaults to `true`.
+      
+      You may want to consider disabling immer when dealing with a large/deep data structure within your state. Immer does unfortunately have an overhead as it wraps data in a Proxy solution for the mutation based updates to work. We would suggest only using this escape hatch if you are noticing any performance degradation.
 
 ## Tutorial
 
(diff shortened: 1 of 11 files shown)


jmyrland:
Hi @s900mhz ! Thanks for looking into this.

With regards to the internal changes, @ctrlplusb has to do a review for this.

In the mean time, the API-docs & TS-types also needs to reflect these changes.

s900mhz:
Sure thing! I'll work on the types and docs tonight.
I am open to any kind of discussion on the API. Just wanted to get this conversation started, we could really use this feature!

website/docs/docs/api/action.md
```diff
@@ -48,6 +48,13 @@ An `action` is a function that is described below.
 
     The payload, if any, that was provided to the
     [action](/docs/api/action.html) when it was dispatched.
+  
+  - `config` (Object)
+
+    The config object that was provided to the
+    [action](/docs/api/action.html) when it was dispatched. 
+    - `immer` (Boolean)
+      Whether to use `immer` to update the state. Defaults to `true`.
```
ctrlplusb, line 57:
```suggestion
    - `immer` (Boolean)
      Whether to use `immer` to update the state. Defaults to `true`.
      
      You may want to consider disabling immer when dealing with a large/deep data structure within your state. Immer does unfortunately have an overhead as it wraps data in a Proxy solution for the mutation based updates to work. We would suggest only using this escape hatch if you are noticing any performance degradation.
```

website/docs/docs/api/reducer.md
```diff
@@ -51,6 +51,12 @@ The `reducer` function is described below.
 
       Any payload that was provided to the action.
 
+  - `config` (Object)
+
+    The config object that was provided to the action when it was dispatched. 
+    - `immer` (Boolean)
+      Whether to use `immer` to update the state. Defaults to `true`.
```
ctrlplusb, line 58:
```suggestion
    - `immer` (Boolean)
      Whether to use `immer` to update the state. Defaults to `true`.
      
      You may want to consider disabling immer when dealing with a large/deep data structure within your state. Immer does unfortunately have an overhead as it wraps data in a Proxy solution for the mutation based updates to work. We would suggest only using this escape hatch if you are noticing any performance degradation.
```

ctrlplusb:
Excellent stuff. Thanks for doing this. Just a few minors and we can get this merged. 🎉

s900mhz:
@ctrlplusb  Awesome! I pushed the changes, lmk what you think!

jmyrland (approved):

jmyrland:
Thanks for your contribution @s900mhz 👏 This will be included in the release this week!

PR #786, opened 2022-10-25
chore: Adds dependabot config

jmyrland:
It seems there is no way to make dependabot ignore the `examples` folder explicitly. See https://github.com/dependabot/dependabot-core/issues/4364

Maybe specifying the root directory ignores the example subfolder 🤷

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+
+updates:
+  - package-ecosystem: npm
+    directory: '/'
+    target-branch: 'master'
+    schedule:
+      interval: 'weekly'
+      day: 'sunday'


no-stack-dub-sack:
@jmyrland [This comment](https://github.com/dependabot/dependabot-core/issues/4364#issuecomment-1118256638) suggests that `directory: '/'` will include all sub-folders. What's your thoughts on switching over to [Renovate](https://github.com/renovatebot/renovate) instead? I've been seeing it on a lot of repos lately and though I don't have any direct experience with it yet, I understand it to be more flexible and configurable than Dependabot, and it looks like it supports this out of the box: https://docs.renovatebot.com/configuration-options/#ignorepaths. It looks like it also has some other features we might benefit from.

jmyrland:
> @jmyrland [This comment](https://github.com/dependabot/dependabot-core/issues/4364#issuecomment-1118256638) suggests that `directory: '/'` will include all sub-folders. What's your thoughts on switching over to [Renovate](https://github.com/renovatebot/renovate) instead? I've been seeing it on a lot of repos lately and though I don't have any direct experience with it yet, I understand it to be more flexible and configurable than Dependabot, and it looks like it supports this out of the box: https://docs.renovatebot.com/configuration-options/#ignorepaths. It looks like it also has some other features we might benefit from.

That was what I was afraid of.
Renovate seems good. Feel free to hijack this PR or create a new one 👍

no-stack-dub-sack:
@jmyrland Sounds good! I'll be away for the better part of the next week or so, but can take this on once I'm back.

jmyrland:
Closing this as it is stale

Issue #765, opened 2022-09-15
Community feedback on `unstable_effectOn`


ctrlplusb:
I'd like to consider moving this API to stable pending feedback from the community.

Also looking to incorporate the work in [PR #598](https://github.com/ctrlplusb/easy-peasy/pull/598)

jmyrland:
Haven't seen this one. What is the usecase for this, over `ActionOn` and `ThunkOn`?

rainerbruemmer:
Hi,

we used the effectOn for serval months and it worked perfect and was very useful. But a a refactoring of our store, we could not use effectOn due to missing [PR #598] (https://github.com/ctrlplusb/easy-peasy/pull/598)

It would be great if the PR will be merged and effectOn gets moved to stable.

ctrlplusb:
> Haven't seen this one. What is the usecase for this, over `ActionOn` and `ThunkOn`?

It allows you to track state rather than actions. Which can help avoid cases of having to track multiple actions in the case of ActionOn and ThunkOn.

jmyrland:
Is #597 still an issue?

methyl:
I think after merging #598 and fixing #597 separately, it will be stable. I didn't notice any other problems than those two and the api itself is really useful.

tormodAase:
Hey, I use `unstable_effectOn` in one of my projects, and it's working great! The only downside with it, is it can only track up to 6 states in the store. I've managed to find some workarounds, but it would be great if it could listen to as many states as we want, similar to how `thunkOn` can listen to unlimited actions.

Thank you so much for maintaining easy-peasy!

ctrlplusb:
Nice one. Yeah, we can look into that quality of life improvement too. 👍

jmyrland:
> The only downside with it, is it can only track up to 6 states in the store. 

This is fixed by #787 👍

ctrlplusb:
@jmyrland and @no-stack-dub-sack on fire. 🔥😎

no-stack-dub-sack:
> I think after merging #598 and fixing #597 separately, it will be stable. I didn't notice any other problems than those two and the api itself is really useful. Thanks @ctrlplusb for finding time for this!

With #598 merged and #597 fixed by #812, I think we're in pretty good shape here. There are no other open issues related to `unstable_effectOn` and no comments in this thread that suggest its not ready to be made stable. 

What do you think @ctrlplusb, @jmyrland?

jmyrland:
[...]

PR #787, opened 2022-10-30
refactor: increase typescript state resolver limit

no-stack-dub-sack:
This PR addresses the [comment](https://github.com/ctrlplusb/easy-peasy/issues/765#issuecomment-1262450643) by @tormodAase in regards to the limit placed upon the numbers of state values that `unstable_effectOn` can track for TypeScript users. As the underlying `StateResolvers` type is shared among `computed` and `unstable_effectOn` this change increases the number of supported state resolvers from 6 to 15. One might argue that anything over 10 is bad design, but there may still be limited use-cases for `effectOn` especially.

It should be noted, however, that if someone does try to exceed this limit, TypeScript simply won't compile without a `@ts-ignore`. There is an alternative approach which supports `N` number of resolvers, but which loses inference for the `Nth + 1` computed func arg or change array member (for `computed` and `effectOn`, respectively). You can find that change (built upon this change) [here](https://github.com/no-stack-dub-sack/easy-peasy/commit/5d4e4d5bf369dd3d49daf9bca93cd664621d48b0). If this is preferred, I can merge these changes and squash them into a single commit.

* update  `StateResolvers` union to support an array up to 15 `StateResolver`s
* update `computed` function's `compFunc` conditional to allow the maximum number of state resolvers
* update `unstable_EffectOn`'s `Dependencies` type to allow the maximum number of state resolvers
* add state-resolvers typescript test to validate that compilation succeeds with up to 15 state resolvers and fails thereafter


jmyrland (commented):
Well done! I'd explore if it is possible to reduce the type definitions by using a dynamic type - but I would have to look more into this to see if it is feasible.

@ctrlplusb can you take a look at this as well?

jmyrland:
If we're going for the dynamic state-resolvers we should probably update the description & maybe review the docs to see if there is any mention of the previous limitation.

As stated in the description, we should advice against using excessive state resolvers (>6?).

no-stack-dub-sack:
> If we're going for the dynamic state-resolvers we should probably update the description & maybe review the docs to see if there is any mention of the previous limitation.
> 
> As stated in the description, we should advice against using excessive state resolvers (>6?).

I agree - I can comb through the docs as well, but I don't think I remember anything about a limit, as this is just a TS limit anyway. So now TypeScript users will just have the same ability as JS users to abuse state resolvers as much as they want :-)

Worth mentioning it would be considered bad practice though.

jmyrland (approved):

PR #788, opened 2022-10-30
test: fix cross-platform persist test issues

no-stack-dub-sack:
Attempt to fix cross-platform issues for persist test `"multiple changes don't cause concurrent persist operations"`. I noticed that this test passed on my mac, but produced highly inconsistent results on my PC. This commit redistributes `wait` times by making updates to the store in more quick succession and waits longer periods to allow the 80ms to elapse before asserting on the store's state. Also make some asserts a bit more specific, asserting that the counter is never intermediate change values, as on Windows after 40-50ms from the last change the counter updates to 5 inconsistently and unexpectedly early. This should prove that the counter never has an intermediate value but eliminate failures due to this inconsistency. We still validate the final value of the counter as 5.

### DIFF
--- a/tests/persist.test.js
+++ b/tests/persist.test.js
@@ -1109,12 +1109,14 @@ test("multiple changes don't cause concurrent persist operations", async () => {
     },
   );
 
+  const DELAY = 10;
+
   // ACT + ASSERTS
   store.getActions().change({
     counter: 1,
   });
 
-  await wait(20);
+  await wait(DELAY);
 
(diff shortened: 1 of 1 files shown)


jmyrland (approved):
LGTM!

no-stack-dub-sack:
@jmyrland did you run this on your machine and confirm it works for you too?

jmyrland:
> @jmyrland did you run this on your machine and confirm it works for you too?

I can verify that it works on my machine as well 👍

Issue #634, opened 2021-03-05
Type Aliases Not Working With 'Thunk' Type


LuisOsta:
**Easy Peasy Version**
```
"easy-peasy": "^5.0.3",
```
**Context**
React SPA Application using CRA. Currently running React 16.9.

So in order to keep my types DRY I tried to define the type for one of my thunks as:
```ts
export type ChangeUserThunk = Thunk<
  EmployeeModel,
  { newRole: USER_ROLE; userId: string },
  Injections,
  IStoreModel
>;
```

**Then use them in my model definition like this:**
```ts
export interface EmployeeModel {
  changeUserRole: ChangeUserThunk;
}
```
**And with my thunk like this:**
```ts
export const changeUserRole: ChangeUserThunk = thunk(
  async (actions, payload, { getStoreState, injections }) => {
    const { newRole, userId } = payload;
    const { userService } = injections;
    const token = getStoreState().user.token;

    try {
      const updatedEmployee = await userService.updateRole(
        userId,
        newRole,
        token
      );
      actions.updateEmployeeById(updatedEmployee);
    } catch (e) {
      console.error(e);
    }
  }
);
```

The problem is when taking this approach the actions, payload, getStoreState, etc are all defined with the default types. So I get these error:
```
Property 'user' does not exist on type 'StateMapper<_Pick<{}, never>>'.ts(2339)
```
```
Property 'updateEmployeeById' does not exist on type 'Actions<{}>'.ts(2339)
```

Notably I also get an error in with the 'ts-toolbet' package. Its as follows:
```
/app/node_modules/ts-toolbelt/out/Any/At.d.ts
client_1  | TypeScript error in /app/node_modules/ts-toolbelt/out/Any/At.d.ts(24,121):
client_1  | Type expected.  TS1110
client_1  | 
client_1  |     22 |  * ```
client_1  |     23 |  */
client_1  |   > 24 | export declare type At<A extends any, K extends Key> = A extends List ? number extends A['length'] ? K extends number | `${number}` ? A[never] | undefined : undefined : K extends keyof A ? A[K] : undefined : unknown extends A ? unknown : K extends keyof A ? A[K] : undefined;
client_1  |        |                                                                                                                         ^
client_1  |     25 |
```

**Important Note**
When switching from the alias method to the spelled out method. It still does not resolve the toolbelt issues.

LuisOsta:
By upgrading I think I was able to solve the toolbelt problem. But the alias thing persists

[...]

PR #791, opened 2022-11-09
Fix thunk and computed alias

arielhs:
This is a followup to [this](https://github.com/ctrlplusb/easy-peasy/pull/790) PR.

Suppose you had a model where you had factored out the `Thunk` and `Computed` types into their own aliases like this:

```
type MyModelAction<TPayload = void> = Action<MyModel, TPayload>
type MyModelComputed<TResult> = Computed<MyModel, TResult>
type MyModelThunk<TPayload = undefined, TResult = any> = Thunk<MyModel, TPayload, any, {}, Promise<TResult>>

type MyState = {
    value: string
}

type MyThunkPayload = { data: string }

interface MyModel {
    myState: MyState

    myComputed: MyModelComputed<number>

    setMyState: MyModelAction<string>

    myThunk: MyModelThunk<MyThunkPayload, boolean>

}

const myModel: MyModel = {
    myState: {
        value: 'initial'
    },

    setMyState: action((state, payload) => {
        state.myState = {
            value: payload
        }
    }),

    myComputed: computed(state => state.myState.value.length),

    myThunk: thunk(async (actions, payload) => {
        actions.setMyState('a')
        // some kind of await
        return true
    })
}
```

Since the `Thunk` and `Computed` types don't actually use the `Model` type parameter in their resulting type, it is "lost" when we make a type alias wrappers. This means the `computed()` and `thunk()` generic functions infer their type parameters like this when attempting to use them on the model instance:

`computed<{}, any, {}, StateResolvers<{}, {}>>`

`thunk<{}, MyThunkPayload, any, {}, Promise<true>>`

Whereas we would expect them to be inferred like this:

`computed<MyModel, number, {}, StateResolvers<MyModel, {}>>`

`thunk<MyModel, MyThunkPayload, any, {}, Promise<true>>`

 The result on the model instance is:

```
// Property 'myState' does not exist on type 'StateMapper<_Pick<{}, never>>'.
myComputed: computed(state => state.myState.value.length),
```

and 

```
//Property 'setMyState' does not exist on type 'Actions<{}>'
actions.setMyState('a')
```

This PR addresses this by adding type markers which forces TypeScript to capture these type variables, fixing the target typing on the model instance.

This has also been used to capture the `StoreModel` and `Injections` type parameters.

A concern I have is if someone was using these types in an unconventional way e.g.

Before:

```
// "type" | "result"
type Before = keyof Computed<any, any>;
```

After:
```
// typeof ModelTypeMarker | typeof StoreModelTypeMarker | "result" | "type"
type After = keyof Computed<any, any>;
```

So in this sense it's a breaking change.

Fixes: https://github.com/ctrlplusb/easy-peasy/issues/634

### DIFF
--- a/tests/typescript/complex-store-inheritance.ts
+++ b/tests/typescript/complex-store-inheritance.ts
@@ -1,0 +1,31 @@
+import { Thunk } from 'easy-peasy';
+
+type Injections = any;
+
+type ISomeOtherModel = {};
+
+interface IModelA<TStore extends ISomeOtherModel> {
+  commitSomething: Thunk<
+    IModelAStoreModel<TStore>,
+    void,
+    Injections,
+    IModelAStoreModel<TStore>
+  >;
+}
+
+// IModelAStoreModel is generic, and requires the root store as input
+type IModelAStoreModel<TStore extends ISomeOtherModel> = IModelA<TStore> &
+  ISomeOtherModel & {
+    onSomethingLoaded: Thunk<TStore, any>;
+  };
+
+// IModelBStoreModel is not generic, and is based on IModelAStoreModel, but it does not care about the generic type
+// of IModelAStoreModel (thus it is just passing`any` to`IModelAStoreModel`)
+interface IModelBStoreModel extends IModelAStoreModel<any> {}
+
+// ❌ Fails here 👇
+// > Interface 'IStoreModel' cannot simultaneously extend types 'IModelAStoreModel<IStoreModel>' and 'IModelBStoreModel'.
+// >   Named property 'commitSomething' of types 'IModelAStoreModel<IStoreModel>' and 'IModelBStoreModel' are not identical.
+interface IStoreModel
+  extends IModelAStoreModel<IStoreModel>,
+    IModelBStoreModel {}
--- a/tests/typescript/complex-store-inheritance.ts
+++ b/tests/typescript/complex-store-inheritance.ts
@@ -23,9 +23,7 @@ type IModelAStoreModel<TStore extends ISomeOtherModel> = IModelA<TStore> &
 // of IModelAStoreModel (thus it is just passing`any` to`IModelAStoreModel`)
 interface IModelBStoreModel extends IModelAStoreModel<any> {}
 
-// ❌ Fails here 👇
-// > Interface 'IStoreModel' cannot simultaneously extend types 'IModelAStoreModel<IStoreModel>' and 'IModelBStoreModel'.
-// >   Named property 'commitSomething' of types 'IModelAStoreModel<IStoreModel>' and 'IModelBStoreModel' are not identical.
+// IStoreModel should be allowed to extend both IModelAStoreModel & IModelBStoreModel
 interface IStoreModel
   extends IModelAStoreModel<IStoreModel>,
     IModelBStoreModel {}
--- a/index.d.ts
+++ b/index.d.ts
@@ -535,7 +535,14 @@ export type Thunk<
   type: 'thunk';
   payload: Payload;
   result: Result;
+  [ModelTypeMarker]?: Model;
+  [InjectionsTypeMarker]?: Injections;
+  [StoreModelTypeMarker]?: StoreModel;
 };
+
+declare const ModelTypeMarker: unique symbol;
+declare const InjectionsTypeMarker: unique symbol;
+declare const StoreModelTypeMarker: unique symbol;
 
 /**
  * Declares an thunk against your model.
@@ -690,6 +697,8 @@ export type Computed<
 > = {
   type: 'computed';
   result: Result;
+  [ModelTypeMarker]?: Model;
+  [StoreModelTypeMarker]?: StoreModel;
 };
 
 type DefaultComputationFunc<Model extends object, Result> = (
--- a/index.d.ts
+++ b/index.d.ts
@@ -525,29 +525,17 @@ type Meta = {
  *   addTodo: Thunk<TodosModel, string>;
  * }
  */
-export type Thunk<
+export interface Thunk<
   Model extends object,
   Payload = undefined,
   Injections = any,
   StoreModel extends object = {},
   Result = any,
-> = {
+> {
   type: 'thunk';
   payload: Payload;
   result: Result;
-  [ModelTypeMarker]?: BivariantCapture<Model>;
-  [InjectionsTypeMarker]?: BivariantCapture<Injections>;
-  [StoreModelTypeMarker]?: BivariantCapture<StoreModel>;
-};
-
-/**
- * Allows the Thunk & Computed type aliases to "capture" the un-used type parameters in a bivariant position
- * so they can't cause any assignability issues, but are still "real" from TypeScript's perspective
- */
-type BivariantCapture<T> = { bivariant(m: T): void }['bivariant'];
-declare const ModelTypeMarker: unique symbol;
-declare const InjectionsTypeMarker: unique symbol;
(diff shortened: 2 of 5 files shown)


jmyrland (commented):
Hi @arielhs ! Thanks for this contribution 👏

I added a typescript test to cover this issue, but I was unable push it to your repository due to insufficient permissions. Could you add this test yourself? (I just copied your example into a `external-type-defs.ts` file in the `tests/typescript` folder)

> This PR addresses this by adding type markers which forces TypeScript to capture these type variables, fixing the target typing on the model instance.

Awesome!

> A concern I have is if someone was using these types in an unconventional way e.g.

Good point. I don't expect this to be heavily utilized, but if so, it is an easy fix for the consumer.

jmyrland:
Sadly, when testing this on a large code base, there are some issues. This does not seem to be working well for composed models.  I'll try to reproduce this in a small example & post back here.

jmyrland:
> Sadly, when testing this on a large code base, there are some issues. This does not seem to be working well for composed models. I'll try to reproduce this in a small example & post back here.

I think I was able to reproduce this - it is atleast one of the issues:

```ts
import { thunk, Thunk } from 'easy-peasy';

type Injections = any;

interface IModelActions<T> {
  someThunk: Thunk<this, T, Injections, IStoreModel>;
}

interface IModel<T> extends IModelActions<T> {
  someValue: T;
  someThunk: Thunk<this, T, Injections, IStoreModel>;
}

interface IStoreModel {
  a: IModel<string>;
  b: IModel<number>;
}

const createModelActions = <T>(): IModelActions<T> => {
  return {
    someThunk: thunk(() => {}),
  };
};

const storeModel: IStoreModel = {
  /*
    Type '{ someThunk: Thunk<IModelActions<string>, string, any, IStoreModel, any>; someValue: string; }' is not assignable to type 'IModel<string>'.
      Types of property 'someThunk' are incompatible.
        Type 'Thunk<IModelActions<string>, string, any, IStoreModel, any>' is not assignable to type 'Thunk<IModel<string>, string, any, IStoreModel, any>'.
          Property 'someValue' is missing in type 'IModelActions<string>' but required in type 'IModel<string>'.ts(2322)

    example.ts(10, 3): 'someValue' is declared here.
    example.ts(15, 3): The expected type comes from property 'a' which is declared here on type 'IStoreModel'
  */
  // ❌ Results in the error above
  a: {
    someValue: 'hello',
    ...createModelActions<string>(),
  },

  // ✅ works fine
  b: {
    someValue: 123,
    someThunk: thunk(() => {}),
  },
};
```

If I remove the marker types in the definition, the problem is resolved:
```ts
export type Thunk<
  Model extends object,
  Payload = undefined,
  Injections = any,
  StoreModel extends object = {},
  Result = any,
> = {
  type: 'thunk';
  payload: Payload;
  result: Result;
  // 👇 When these are "removed", the issue above dissappears 🤔
  // [ModelTypeMarker]?: Model;
  // [InjectionsTypeMarker]?: Injections;
  // [StoreModelTypeMarker]?: StoreModel;
};
```

I'm not sure how to resolve this issue. We should be able to compose stores, as the example above with the `a`-store.

arielhs:
@jmyrland I'll look at the example now and see if i can find another way that works. I've also added you as a collaborator to my fork

arielhs:
@jmyrland I found a way to capture the type parameters without creating the assignability issues that would break the composed models like you described in your example. I've just pushed, let me know if this works against the real example - the large codebase you described.

I am confident it is a safe technique, see the "Methodish" example [here](https://www.typescriptlang.org/tsconfig#strictFunctionTypes) if you are interested

jmyrland:
> @jmyrland I found a way to capture the type parameters without creating the assignability issues that would break the composed models like you described in your example. I've just pushed, let me know if this works against the real example - the large codebase you described.
> 
> I am confident it is a safe technique, see the "Methodish" example [here](https://www.typescriptlang.org/tsconfig#strictFunctionTypes) if you are interested

Awesome! You're a TS wizard! 😁 

That fixed the specific issue - but this is whack-a-mole, and another one popped up (generalized output):

> Interface 'IStoreModel' cannot simultaneously extend types 'IModelA<IStoreModel, IModelB<IStoreModel, any>>' and 'IModelC'.
  Named property 'someAction' of types IModelA<IStoreModel, IModelB<IStoreModel, any>>' and 'IModelC' are not identical.

I'll see if I can construct a test to reproduce this issue, unless you immediately see the fix.

arielhs:
> > @jmyrland I found a way to capture the type parameters without creating the assignability issues that would break the composed models like you described in your example. I've just pushed, let me know if this works against the real example - the large codebase you described.
> > I am confident it is a safe technique, see the "Methodish" example [here](https://www.typescriptlang.org/tsconfig#strictFunctionTypes) if you are interested
> 
> Awesome! You're a TS wizard! 😁
> 
> That fixed the specific issue - but this is whack-a-mole, and another one popped up (generalized output):
> 
> > Interface 'IStoreModel' cannot simultaneously extend types 'IModelA<IStoreModel, IModelB<IStoreModel, any>>' and 'IModelC'.
> > Named property 'someAction' of types IModelA<IStoreModel, IModelB<IStoreModel, any>>' and 'IModelC' are not identical.
> 
> I'll see if I can construct a test to reproduce this issue, unless you immediately see the fix.

Lol I'm happy to keep whack-a-moling, please do make the example

jmyrland:
Elegant solution @arielhs 👏

I've tested the new types against the "Large Codebase ™️" and all is good ✅

arielhs:
> > Lol I'm happy to keep whack-a-moling, please do make the example
> 
> I've added a new testcase for complex store inheritance, extrapolated from existing code. It is quite convoluted - but if this passes, so should the existing code.

Thanks again for creating an example. I feel a bit silly, the solution ended up being much simpler... just change `Thunk` and `Computed` to interfaces! It actually makes sense because the whole cause of this issue is that newer versions of TypeScript are more aggressive at simplifying type aliases, but interfaces are never simplified away.

jmyrland (dismissed):

jmyrland:
> Thanks again for creating an example. I feel a bit silly, the solution ended up being much simpler... just change `Thunk` and `Computed` to interfaces! It actually makes sense because the whole cause of this issue is that newer versions of TypeScript are more aggressive at simplifying type aliases, but interfaces are never simplified away.

My pleasure - I very much appreciate the simple solution of using `interface` 👍

Let's get this released when @ctrlplusb has a moment to spare! 🚀

tests/typescript/complex-store-inheritance.ts
```diff
@@ -0,0 +1,31 @@
...
+  };
+
+// IModelBStoreModel is not generic, and is based on IModelAStoreModel, but it does not care about the generic type
+// of IModelAStoreModel (thus it is just passing`any` to`IModelAStoreModel`)
+interface IModelBStoreModel extends IModelAStoreModel<any> {}
+
+// ❌ Fails here 👇
+// > Interface 'IStoreModel' cannot simultaneously extend types 'IModelAStoreModel<IStoreModel>' and 'IModelBStoreModel'.
+// >   Named property 'commitSomething' of types 'IModelAStoreModel<IStoreModel>' and 'IModelBStoreModel' are not identical.
+interface IStoreModel
```
jmyrland, line 29:
```suggestion
// IStoreModel should be allowed to extend both IModelAStoreModel & IModelBStoreModel
interface IStoreModel
```

jmyrland (approved):

jmyrland:
Thanks again for your contribution @arielhs ! 👏 This will be included in the upcoming release this week.

PR #803, opened 2022-12-15
fix: allow `Unstable_EffectOn` type to be aliased

### LINKED ISSUE #802: Unable to alias `Unstable_EffectOn` type
Similar to #634 and the issue described and fixed in #790, users of easy-peasy were unable to define their own aliases of the `Action`, `Computed`, and `Thunk` types. The same is true for `Unstable_EffectOn`, e.g. 

```ts
type AppEffectOn<TModel extends object> = Unstable_EffectOn<
  TModel,
  StoreModel,
  Injections
>;

interface StoreModel {
  todos: TodosModel;
  rootData: number;
}

interface TodosModel {
  items: { id: number; text: string }[];
  foo: string;
  setFoo: Action<TodosModel, string>;
  onStateChanged: AppEffectOn<TodosModel>;
}

const todosModel: TodosModel = {
  // ...
  onStateChanged: unstable_effectOn(
    [
      (state) => state.items, // error! Property 'items' does not exist on type 'StateMapper<FilterActionTypes<{}>>'
      (state, storeState) => storeState.rootData, // error!
    ],
    (actions, change, helpers) => {
      actions.setFoo('bar'); // error! Property 'setFoo' does not exist on type 'Actions<{}>'
      helpers.dispatch.todos.setFoo('plop'); // error!
      helpers.getState().items[0].id + 2; // error!

      return () => console.log('dispose');
    },
  ),
};

Since the generic type params of the type are not actually used, TypeScript "forgets" them when the type is aliased. The fix should be straightforward which I can spin up a PR for shortly.

no-stack-dub-sack:
- Following up on the work completed in #790 and #791, defines the `Unstable_EffectOn` type as an interface rather than a type so that unused type params don't get "lost" when the type is aliased. 
- Adds/updates TypeScript tests

fixes #802

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -698,13 +698,13 @@ export function computed<
 
 // #region EffectOn
 
-export type Unstable_EffectOn<
+export interface Unstable_EffectOn<
   Model extends object = {},
   StoreModel extends object = {},
   Injections = any,
-> = {
+> {
   type: 'effectOn';
-};
+}
 
 type Change<Resolvers extends StateResolvers<any, any>> = {
   prev: ExtractReturnTypes<Resolvers>;
--- a/tests/typescript/external-type-defs.ts
+++ b/tests/typescript/external-type-defs.ts
@@ -1,5 +1,15 @@
-import { action, Action, computed, Computed, thunk, Thunk } from 'easy-peasy';
+import {
+  action,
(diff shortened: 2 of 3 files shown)


jmyrland (approved):
LGTM 👍

PR #810, opened 2023-01-14
fix: types and docs related to action config obj

no-stack-dub-sack:
**This PR:**
- Is related to the work completed in #781 and is mainly to update docs and types related to configuring actions / disabling Immer. The config object is the second argument to `action` and `reducer` and both the types and docs needed to be updated to reflect this.
- Adds an example in the action docs for disabling Immer to make it clearer immutability needs to be managed manually when `immer` is set to `false`. ([preview docs](https://easy-peasy-git-fix-immer-config-types-docs-ctrlplusb.vercel.app/docs/api/action.html#disabling-immer)).
- Fixes a fall-through case in the `reducer` example in the docs
- Updates a few tests related to disabling Immer.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -593,9 +593,8 @@ export type Action<Model extends object, Payload = void> = {
   result: void | State<Model>;
 };
 
-
 /**
-  * @param {boolean} [immer=true] - If true, the action will be wrapped in an immer produce call. Otherwise, the action will update the state directly.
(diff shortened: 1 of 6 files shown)


jmyrland (dismissed):
👍

no-stack-dub-sack:
@jmyrland can you please review the last commit and re-approve when you get a chance? I noticed a typo when reviewing the preview deployment and just fixed. Thanks!

jmyrland (approved):
LGTM 👍

PR #816, opened 2023-02-03
feat: Pass full state to persist transforms

damassi:
Noticed a slight divergence from the `redux-persist` [transform api](https://github.com/rt2zz/redux-persist/blob/master/src/createTransform.ts#L26-L33), where there they pass the full state along with the `in` and `out` calls. This also aligns with the api as seen in `create-transform` code in the easy-peasy [codebase](https://github.com/ctrlplusb/easy-peasy/blob/master/src/create-transform.js#L17-L24). 

Now the api is: 
```tsx
in: (state, key, fullState) => { ... }
out: (state, key, fullState) => { ... }
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1032,8 +1032,8 @@ export interface PersistStorage {
 }
 
 export interface Transformer {
-  in?: (data: any, key: string) => any;
-  out?: (data: any, key: string) => any;
+  in?: (data: any, key: string, fullState?: any) => any;
+  out?: (data: any, key: string, fullState?: any) => any;
 }
 
 export interface PersistConfig<Model extends object> {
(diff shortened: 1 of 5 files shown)


no-stack-dub-sack (changes requested):
@damassi this looks good to me - would you mind taking a look at the persist tests and seeing if it makes sense to add or modify any existing tests to reflect the new API?

damassi:
Sure thing, I'll do this now.

damassi:
@no-stack-dub-sack Updated tests to reflect the additional parameter being passed into the transform. Let me know if this is sufficient!

no-stack-dub-sack (approved):

no-stack-dub-sack:
Thanks @damassi! LGTM!

PR #818, opened 2023-02-04
feat: Add persisted state migrations

damassi:
Similar to [`redux-persist`](https://github.com/rt2zz/redux-persist#migrations), this PR adds the ability to migrate persisted state using a new `migrations` option on the `persist` helper. It looks like this: 

```tsx
const memoryStorage = createMemoryStorage({
  '[EasyPeasyStore][0]': {
    foo: "foo-updated",
    migrationConfict: "error"
  },
});

const store = createStore(
  persist(
    {
      foo: {
        bar: "bar",
      },
      bar: "bar",
    },
    {
      storage: memoryStorage,
      migrations: {
        migrationVersion: 1,

        0: (state) => {
          state.foo = { bar: state.foo };
        },
        1: (state) => {
          delete state.migrationConfict;
        },
      },
    }
  )
);
```

Yielding this updated store representation:
```json
{ 
  "foo": { 
    "bar": "foo-updated" 
  }, 
  "bar": "bar", 
  "_migrationVersion": 1 
}
```

**How it works**: 
- Numeric object keys are added to the `migrations` object
- Each migration function is an immer draft updator, which takes state
- The migrator will look at the currently stored internal version (`state._migrationVersion`) and perform migrations up to `migrations.migrationVersion`, as defined in the store configuration setup. 

**Questions**:
1. We need to persist the currently migrated version, as that value is what's compared against for future migrations. Right now it's been added as `_migrationVersion` but perhaps there's a better place for this? 
2. Anything I might be missing in the implementation? It felt surprisingly straightforward.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1040,6 +1040,10 @@ export interface PersistConfig<Model extends object> {
   allow?: Array<keyof Model>;
   deny?: Array<keyof Model>;
   mergeStrategy?: 'mergeDeep' | 'mergeShallow' | 'overwrite';
+  migrations?: {
+    migrationVersion: number;
+    [key: number]: (state: Partial<Model & { [key: string | number]: any }>) => void;
+  }
   storage?: 'localStorage' | 'sessionStorage' | PersistStorage;
   transformers?: Array<Transformer>;
 }
--- a/src/migrations.js
+++ b/src/migrations.js
@@ -6,7 +6,7 @@ export const migrate = (
 ) => {
   setAutoFreeze(false);
 
-  let version = data._migrationVersion ?? -1;
+  let version = data._migrationVersion ?? 0;
   const toVersion = migrations.migrationVersion
 
   if (typeof version !== "number" || typeof toVersion !== 'number') {
--- a/src/migrations.js
+++ b/src/migrations.js
@@ -6,10 +6,10 @@ export const migrate = (
 ) => {
   setAutoFreeze(false);
 
-  let version = data._migrationVersion || -1;
+  let version = data._migrationVersion ?? -1;
   const toVersion = migrations.migrationVersion
 
-  if (typeof toVersion !== 'number') {
+  if (typeof version !== "number" || typeof toVersion !== 'number') {
     throw new Error('No migration version found');
   }
 
--- a/src/migrations.js
+++ b/src/migrations.js
@@ -1,0 +1,31 @@
+import { produce, setAutoFreeze } from 'immer';
+
+export const migrate = (
+  data,
+  migrations,
+) => {
+  setAutoFreeze(false);
+
+  let version = data._migrationVersion || -1;
+  const toVersion = migrations.migrationVersion
+
+  if (typeof toVersion !== 'number') {
+    throw new Error('No migration version found');
+  }
+
+  while (version < toVersion) {
+    const nextVersion = version + 1;
+    const migrator = migrations[nextVersion];
+
+    if (!migrator) {
+      throw new Error(`No migrator found for \`migrationVersion\` ${nextVersion}`);
+    }
+
+    data = produce(data, migrator);
+    data._migrationVersion = nextVersion;
+    version = data._migrationVersion;
+  }
+
+  setAutoFreeze(true);
+  return data;
+}
--- a/src/persistence.js
+++ b/src/persistence.js
@@ -1,4 +1,5 @@
 import { clone, get, isPlainObject, isPromise, set, pSeries } from './lib';
+import { migrate } from './migrations';
 
 const noopStorage = {
   getItem: () => undefined,
@@ -32,7 +33,7 @@ const getBrowerStorage = (storageName) => {
 const localStorage = getBrowerStorage('localStorage');
 const sessionStorage = getBrowerStorage('sessionStorage');
 
-function createStorageWrapper(storage, transformers = []) {
+function createStorageWrapper(storage, transformers = [], migrations = {}) {
   if (storage == null) {
     storage = sessionStorage();
   }
@@ -69,10 +70,14 @@ function createStorageWrapper(storage, transformers = []) {
   };
 
   const deserialize = (data) => {
-    const result =
+    const storageData =
       storage === localStorage() || storage === sessionStorage()
         ? JSON.parse(data).data
         : data;
+
+    const hasMigrations = Object.keys(migrations).length > 0;
+    const result = hasMigrations ? migrate(storageData, migrations) : storageData;
+
     if (
       outTransformers.length > 0 &&
       result != null &&
@@ -117,6 +122,7 @@ export function extractPersistConfig(path, persistdef = {}) {
       storage: createStorageWrapper(
         persistdef.storage,
         persistdef.transformers,
+        persistdef.migrations,
       ),
     },
   };
--- a/tests/persist.test.js
+++ b/tests/persist.test.js
@@ -821,6 +821,56 @@ test('migrations', async () => {
   expect(store.getState().userSession).toBeUndefined();
 })
 
+test('partially applied migrations', async () => {
+  // ARRANGE
+  const memoryStorage = createMemoryStorage({
+    '[EasyPeasyStore][0]': {
+      user: {
+        name: "User Name",
+      },
+      userSession: "session",
+      _migrationVersion: 1,
+    },
+  });
+
+  const makeStore = () =>
+    createStore(
+      persist(
+        {
+          user: {
+            name: null,
+            session: null
+          },
+        },
+        {
+          storage: memoryStorage,
+          migrations: {
+            migrationVersion: 2,
+
+            1: (state) => {
+              state.user = { name: state.user }
+              state.userSession = state.session;
+              delete state.session;
+            },
+            2: (state) => {
+              state.user.session = state.userSession;
+              delete state.userSession;
+            },
+          }
+        },
+      ),
+    );
+
+  const store = makeStore();
+  await store.persist.resolveRehydration();
+
+  // ASSERT
+  expect(store.getState().user.name).toBe('User Name')
+  expect(store.getState().user.session).toBe('session')
+  expect(store.getState().session).toBeUndefined();
+  expect(store.getState().userSession).toBeUndefined();
+})
+
 test('multiple stores', async () => {
   // ARRANGE
   const memoryStorage = createMemoryStorage();
--- a/tests/persist.test.js
+++ b/tests/persist.test.js
@@ -774,6 +774,48 @@ test('transformers order', async () => {
   });
 });
 
+test('migrations', async () => {
+  // ARRANGE
+  const memoryStorage = createMemoryStorage({
+    '[EasyPeasyStore][0]': {
+      foo: "foo-updated",
+      migrationConfict: "error"
+    },
+  });
+
+  const makeStore = () =>
+    createStore(
+      persist(
+        {
+          foo: {
+            bar: "bar",
+          },
+          bar: "bar",
+        },
+        {
+          storage: memoryStorage,
+          migrations: {
+            migrationVersion: 1,
+
+            0: (state) => {
+              state.foo = { bar: state.foo }
+            },
+            1: (state) => {
+              delete state.migrationConfict;
+            },
+          }
+        },
+      ),
+    );
+
+  const store = makeStore();
+  await store.persist.resolveRehydration();
+
+  // ASSERT
+  expect(store.getState().foo.bar).toBe('foo-updated')
+  expect(store.getState().migrationConfict).toBeUndefined();
+})
+
 test('multiple stores', async () => {
   // ARRANGE
   const memoryStorage = createMemoryStorage();
--- a/tests/persist.test.js
+++ b/tests/persist.test.js
@@ -778,8 +778,8 @@ test('migrations', async () => {
   // ARRANGE
   const memoryStorage = createMemoryStorage({
     '[EasyPeasyStore][0]': {
-      foo: "foo-updated",
-      migrationConfict: "error"
+      session: "session",
+      user: "User Name",
     },
   });
 
@@ -787,10 +787,10 @@ test('migrations', async () => {
     createStore(
       persist(
         {
-          foo: {
-            bar: "bar",
+          user: {
+            name: null,
+            session: null
           },
-          bar: "bar",
         },
         {
           storage: memoryStorage,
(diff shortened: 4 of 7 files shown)


tests/persist.test.js
```diff
@@ -774,6 +774,48 @@ test('transformers order', async () => {
...
+        {
+          foo: {
+            bar: "bar",
+          },
+          bar: "bar",
+        },
+        {
+          storage: memoryStorage,
+          migrations: {
+            migrationVersion: 1,
```
damassi, line 798: Didn't want to use `version` here so as to not confuse the top-level `persistKey` version. `migrationVersion` seems sufficiently unambiguous

src/migrations.js
```diff
@@ -0,0 +1,31 @@
+import { produce, setAutoFreeze } from 'immer';
+
+export const migrate = (
+  data,
+  migrations,
+) => {
+  setAutoFreeze(false);
+
+  let version = data._migrationVersion || -1;
```
damassi, line 9: Grab the last migration that has run, and if we've never run a migration before set an initial value

src/migrations.js
```diff
@@ -0,0 +1,31 @@
...
+  while (version < toVersion) {
+    const nextVersion = version + 1;
+    const migrator = migrations[nextVersion];
+
+    if (!migrator) {
+      throw new Error(`No migrator found for \`migrationVersion\` ${nextVersion}`);
+    }
+
+    data = produce(data, migrator);
+    data._migrationVersion = nextVersion;
```
damassi, line 25: Save the version

src/migrations.js
```diff
@@ -0,0 +1,31 @@
...
+    version = data._migrationVersion;
+  }
+
+  setAutoFreeze(true);
```
damassi, line 29: Turn immer's object freezing capabilities back on now that we've set the latest `migrationVersion` and are done migrating

src/persistence.js
```diff
@@ -69,10 +70,14 @@ function createStorageWrapper(storage, transformers = []) {
...
 
   const deserialize = (data) => {
-    const result =
+    const storageData =
       storage === localStorage() || storage === sessionStorage()
         ? JSON.parse(data).data
         : data;
+
+    const hasMigrations = Object.keys(migrations).length > 0;
+    const result = hasMigrations ? migrate(storageData, migrations) : storageData;
```
damassi, line 79: It makes sense to run migrations before additional transformations occur since migrations are structural, where transformations operate on the leafs

no-stack-dub-sack:
Thanks for the PR! This is a cool concept which solves for what seems like a pretty common use-case. We will take a closer look at your changes / questions, and since this introduces a completely new feature, and since he's the most familiar with the persist implementation, I think we'll need @ctrlplusb to take a look and give his final sign off. If accepted, it would be nice to get this in before the upcoming 6.0.0 release which fixes a few minor issues and stabilizes the `unstable_effectOn` api.

damassi:
Amazing. We've been using our own version of this on top of easy-peasy for a few years in our mobile app and its been working great but felt it would be better to simplify things by PRing upstream. Thanks for giving it a look 👍

damassi:
Hi @ctrlplusb - any chance you would have a moment to give this PR a review?

no-stack-dub-sack (changes requested):
@damassi it looks like we're moving towards accepting this feature, likely to include as part of the upcoming v6 release which stabilizes `unstable__effectOn`.  Apologies for the delay on this. 

Before accepting this, though, we'd want to round it out with some docs and perhaps some more exhaustive tests. Is this something you'd be up for taking on?

As far as `_migrationVersion`, I think this is OK for now, but I'll take another look as well and see if there's maybe a better place for this.

Thanks!

damassi:
I can certainly add docs! Let me think about further test-cases; as the underlying implementation is fundamentally quite simple, there's not much to it but surely I'm missing something. I'll try to wrap this up today / tomorrow.

index.d.ts
```diff
@@ -1040,6 +1040,10 @@ export interface PersistConfig<Model extends object> {
   allow?: Array<keyof Model>;
   deny?: Array<keyof Model>;
   mergeStrategy?: 'mergeDeep' | 'mergeShallow' | 'overwrite';
+  migrations?: {
+    migrationVersion: number;
+    [key: number]: (state: Partial<Model & { [key: string | number]: any }>) => void;
```
damassi, line 1045: In order to support `delete` operations in migrator functions, make state `Partial`.

damassi:
@no-stack-dub-sack - This is now ready for review.

website/docs/docs/api/persist.md
```diff
@@ -69,9 +71,26 @@ Below is the API of the `persist` helper function.
...
+  - `migrations` (Object, _optional_)
+
+    Are used to transform persisted store state from one representation to
+    another. It contains the following properties:
+
+    - `migrationVersion` (number, _required_)
+
+      The migration version to execute up to, starting at 0.
+
+    - `[0..1..n]` (Function, _required_)
```
damassi, line 83: @no-stack-dub-sack I'm having trouble figuring out how to express this prop name in the API docs... The description below explains whats going on, but any suggestions as to how to write it here?

src/migrations.js
```diff
@@ -0,0 +1,31 @@
+import { produce, setAutoFreeze } from 'immer';
+
+export const migrate = (
+  data,
+  migrations,
+) => {
+  setAutoFreeze(false);
+
+  let version = data._migrationVersion ?? 0;
```
damassi, line 9: Q: Should we start migration versions from `0` or `1`? Meaning, a user adds their first migration:

```ts
migrations: {
  migrationVersion: 0,
  
  0: (state) => { 
    ...
  }
}
```
Or...

```ts
migrations: {
  migrationVersion: 1,
  
  1: (state) => { 
    ...
  }
}
```

Since we're not using an array here I rather like the latter option, but perhaps it could be confusing. Thoughts?

website/docs/docs/api/persist.md
jmyrland: I get your point - reading this might be a little confusing.

Maybe we could illustrate the migration config - with supplying texts instead?

> - `migrations` (Object, _optional_)
> 
> This config is used to transform persisted store state from one representation to another. This object is keyed by version numbers, with migration functions attached to each version. A `migrationVersion` is also required for this object, to specify which version to target. 
> ```js
> persist(
>   {...},
>   {
>     migrations: {
>       migrationVersion: 2,
> 
>       1: (state) => { ... },
>       2: (state) => { ... },
>     },
>   }
> );
> ```

A little more context might make it easier to "get"?

jmyrland (commented):
This is looking good @damassi 👏

website/docs/docs/api/persist.md
damassi: Updated!

src/migrations.js
no-stack-dub-sack: Likewise

damassi: Cool, will resolve

website/docs/docs/api/persist.md
```diff
@@ -246,16 +265,84 @@ have persisted state based on a previous version of your store model. The user's
...
+```
+
+### Forced updates via `version`
+
+If migrations are insufficient (which can often be the case after a major state
+refactor has taken place), the persist API also provides a means to "force
+update" the store via `version`.
+
+By default the persist API utilizes the `mergeDeep` strategy (you can read more
+above merge strategies further below). The `mergeDeep` strategy attempts to
```
no-stack-dub-sack, line 338: This looks like an old typo but we might as well fix it: 

```suggestion
about merge strategies further below). The `mergeDeep` strategy attempts to
```

no-stack-dub-sack (dismissed):
This is looking solid @damassi! The docs look great (the example is very clear and helpful) and the tests are all passing. One small typo that's not even yours, so that's optional. Let's let @jmyrland have a final weigh in here as well but this is looking good from my end. 

Nice work!

no-stack-dub-sack:
@damassi @jmyrland On second thought about the docs, what do you think about moving the section about the default behavior up to right under the "Managing Model Updates" header, then introducing the two options, so it could read something like:

<hr />

## Managing model updates
It's entirely reasonable that your store model will evolve over time. However, there is an inherent risk with this when utilizing the persist API.

If your application has previously been deployed to production then users may have persisted state based on a previous version of your store model. The user's persisted state may not align with that of your new store model, which could result in errors when a component tries to consume/update the store.

By default the persist API utilizes the mergeDeep strategy (you can read more above merge strategies further below). The mergeDeep strategy attempts to perform an optimistic merge of the persisted state against the store model. Where it finds that the persisted state is missing keys that are present in the store model, it will ensure to use the respective state from the store model. It will also verify the types of data at each key. If there is a misalignment (ignoring null or undefined) then it will opt for using the data from the store model instead as this generally indicates that the respective state has been refactored.

Whilst the mergeDeep strategy is fairly robust and should be able to cater for a wide variety of model updates, it can't provide a 100% guarantee that a valid state structure will be resolved.

Therefore, when dealing with production applications, we recommend that you consider removing this risk by using one of the two options described below:

### Migrations
`<your docs here>`

### Forced updates via `version`
If migrations are insufficient (which can often be the case after a major state refactor has taken place), the persist API also provides a means to "force update" the store via version. You can do so by utilizing the version configuration property that is available on the store config.

`<original docs here>`

damassi:
Yup, I agree that this reordering works much better. Updated 👍

website/docs/docs/api/persist.md
```diff
@@ -246,23 +268,88 @@ have persisted state based on a previous version of your store model. The user's
...
-`null` or `undefined`) then it will opt for using the data from the store model
-instead as this generally indicates that the respective state has been
+By default the persist API utilizes the mergeDeep strategy (you can read more
+about merge strategies further below). The mergeDeep strategy attempts to
+perform an optimistic merge of the persisted state against the store model.
+Where it finds that the persisted state is missing keys that are present in the
+store model, it will ensure to use the respective state from the store model. It
+will also verify the types of data at each key. If there is a misalignment
+(ignoring null or undefined) then it will opt for using the data from the store
+model instead as this generally indicates that the respective state has been
```
jmyrland, line 278:
```suggestion
By default the persist API utilizes the `mergeDeep` strategy (you can read more
about merge strategies further below). The `mergeDeep` strategy attempts to
perform an optimistic merge of the persisted state against the store model.
Where it finds that the persisted state is missing keys that are present in the
store model, it will ensure to use the respective state from the store model. It
will also verify the types of data at each key. If there is a misalignment
(ignoring `null` or `undefined`) then it will opt for using the data from the store
model instead as this generally indicates that the respective state has been
```

I think we still want code highlights here?

website/docs/docs/api/persist.md
```diff
@@ -246,23 +268,88 @@ have persisted state based on a previous version of your store model. The user's
...
 refactored.
 
-Whilst the `mergeDeep` strategy is fairly robust and should be able to cater for
-a wide variety of model updates, it can't provide a 100% guarantee that a valid
+Whilst the mergeDeep strategy is fairly robust and should be able to cater for a
```
jmyrland, line 281:
```suggestion
Whilst the `mergeDeep` strategy is fairly robust and should be able to cater for a
```
And here?

jmyrland (approved):
Some minor highlights missing, other than that - LGTM 👍

website/docs/docs/api/persist.md
no-stack-dub-sack: I think these were probably lost when I copied and pasted into the GH comment. The order I suggested was probably copied from there rather than re-ordered from the docs source itself. Looks good now in any case!

damassi: Yup I had copied that over from the comment; all fixed now

no-stack-dub-sack:
@damassi @jmyrland I started thinking a bit more about how migrations and merge strategies may/may not work together and came up with this scenario. Let's say you have the following persisted state:

```ts
{
  city: 'london',
  postCode: 'e3 1pq',
  id: 1,
}
```
and you add a migration to update the model to the following type:

```ts
type State = {
  address: {
    city: string;
    postCode: string;
  };
  id: string;
}
```
However the migration you wrote only does this:
```ts
{
  1:  (state) => {
    state.address = {
      city: state.city,
      postCode: state.postCode,
    }

    delete state.city;
    delete state.postCode;
  }
}
```
But neglects to add `state.id = String(state.id)`.

In this case, the structural change will be applied, but there will be a data type mismatch between `id` in the user's persisted state and `id` from the model, and since merge strategies are applied after migrations, the value from the store model will be used due to the mismatch.

I was originally thinking that merge strategies being applied after migrations would pretty much be a no-op, but I believe this example illustrates a use-case where both migrations _and_ merge strategies will impact the resolved state (noting, however, that a better-written migration _would have_ made the merge strategy a no-op).

I don't necessarily see this as being an issue, but do you think its worth calling out that merge strategies _may_ work in conjunction with migrations, and that they're not always mutually exclusive?

jmyrland:
> As long as users writing migrations appreciate the risk (which is well outlined in the current docs, I think) it will prompt them to take the migration writing process seriously. 

Makes sense - it's in the hands of the developer at that stage.

> If this can be emphasized further, however, all for it.

A short warning at the end of the migrations section might be warranted?

```markdown
⚠ Whenever properties in your store changes types, ensure that you migrate these properly. 
Persisted data might be lost in when [merging](#merge-strategies), if there is a mismatch 
between types in the store & the persisted data (excluding `null` and `undefined`).
```

Maybe another short paragraph, to encourage writing tests for migrations to ensure correctness?

These additions might be overkill, and lead to more confusion/bloated docs. I'm guessing that people in need of migrations will test this properly before "release".

Thoughts @no-stack-dub-sack @damassi ?

damassi:
I would vote for skipping since the persist docs are already quite long and descriptive. If one is at the point in their apps life-cycle where they're writing migrations then these issues will naturally be apparent (imo). More than happy to add whatever y'all think is needed though.

no-stack-dub-sack:
@damassi I think I agree. I would either vote for skipping it or a simpler two-liner. Something like:

> Well-written migrations should obviate the need for merge strategies and render them no-ops. Note, however, that merge strategies are still applied and unexpected behavior may occur during rehydration as a result of incomplete migrations.

If either of you think that this may still invite confusion, I'm down to skip it since as @damassi said, developers at this stage in their app development should generally be aware of these issues and its on them to handle them correctly.

damassi:
I actually quite like your suggestion @no-stack-dub-sack; I think that puts it well in just the right amount of words. If you'd like to proceed with it, let me know where exactly to drop it into the doc.

no-stack-dub-sack:
@damassi I think it can go right before the example, as a new paragraph, or right after. It might make more sense after, but its possible it could also be missed there. I'd be ok with either place, though, so whatever you think makes the most sense.

damassi:
Moved it below, where it reads / flows a bit better.

no-stack-dub-sack:
Great work here @damassi! This is looking good. I've added one more test and I'm merging this in. @jmyrland and I will work on prepping a v6 release very soon 👍

no-stack-dub-sack (approved):
LGTM!

damassi:
Thanks all! Very excited to pull out all of our homegrown code and drop this into our apps, and simplify. We <3 easy peasy over at Artsy.

PR #823, opened 2023-03-30
bugfix: Nextjs SSR issues (fixes #822)

### LINKED ISSUE #822: Broken in NextJS 13 with React 18
I'm getting an error with `TypeError: maybeGetServerSnapshot is not a function` when rendering a component that calls useStoreState:

```
TypeError: maybeGetServerSnapshot is not a function
This error happened while generating the page. Any console logs will be displayed in the terminal window.
Source
getServerSnapshotWithSelector
file:/Users/feral/Projects/trendradar/node_modules/use-sync-external-store/cjs/use-sync-external-store-shim/with-selector.development.js (138:31)
Object.useSyncExternalStore
file:/Users/feral/Projects/trendradar/node_modules/react-dom/cjs/react-dom-server.browser.development.js (5217:10)
useSyncExternalStore
file:/Users/feral/Projects/trendradar/node_modules/react/cjs/react.development.js (1676:21)
useSyncExternalStoreWithSelector
file:/Users/feral/Projects/trendradar/node_modules/use-sync-external-store/cjs/use-sync-external-store-shim/with-selector.development.js (145:15)
useStoreState
file:/Users/feral/Projects/trendradar/node_modules/easy-peasy/dist/index.cjs.js (66:25)
```

Using the latest React/NextJS and easy-peasy in Node v18.15.0

jmyrland:
It seems that this issue is spawned by the `useSyncExternalStoreWithSelector` is initialized with `getServerSnapshot` set to an object, when a function is expected - because the `getServerState` (from `createStore()`) returns the initial state directly.

I've refactored this, to set the `getServerSnapshot` equal to the `getState` method from the store.
As I no longer see the need for the `getServerState` (from `createStore()`), I've removed it.

- Bumps nextjs example to latest version (does not reproduce the issue)
- Adds a `nextjs-ssr` example (reproduces the issue)
- Extends server render tests

related to #822


no-stack-dub-sack (approved):
I didn't dig deep into reviewing the examples, but the core code change makes sense and looks good to me.

PR #824, opened 2023-04-01
feat: React Native Todo

bahaaador:
- React Native todo app init
- Copy store from simple todo app
- Add Todo and Toolbar components
- Add readme and updated examples Readme


jmyrland (approved):
Thanks for your contribution 👏 The android version wokrs as well 🎉

PR #836, opened 2023-06-12
[Bugfix] persist not working on iOS

jmyrland:
`createPersistor` defines a `timingMethod` that won't work on iOS.

This is because this method utilizes `requestIdleCallback`, which is defined on iOS - but never invoked 🤯

So, for iOS we have to fallback to utilize `requestAnimationFrame`. This is done by vendoring a specific file `timingMethod.ios.js`, which the RN bundle can import for the iOS platform.

### DIFF
--- a/src/persistence.js
+++ b/src/persistence.js
@@ -1,3 +1,4 @@
+import timingMethod from './utils/timingMethod';
 import { clone, get, isPlainObject, isPromise, set, pSeries } from './lib';
 import { migrate } from './migrations';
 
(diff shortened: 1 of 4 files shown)


jmyrland:
I'm not able to verify if this is working, as I currently don't have a mac to test this on :/

jmyrland:
Verified and tested on https://snack.expo.dev/_fo68oUf7

Notice how `SET_ITEM` is logged for all platforms.

no-stack-dub-sack (approved):
LGTM!

no-stack-dub-sack (approved):

PR #847, opened 2023-07-14
chore: Bump dependencies

jmyrland:
Upgrades all dependencies to their respective latest version.

Ignored updates:
- `rollup`: due to build issues
- `immer`: due to breaking changes

All internal tests is confirmed to pass locally.
I've also tested the updated version against a huge project that is heavily utilizing `easy-peasy` with over 2k unit tests, all passing.


jmyrland:
> @jmyrland Seeing such a large diff in the package.json admittedly makes me a little nervous (though most of them are dev deps which I'm less concerned about), but sounds like you've tested it very well. So I assume none of the breaking Immer changes affect us here with how we use immer? Do we know what those changes were and is it possible our tests may not cover them?

I agree - but having backing tests helps alot! However, my external tests only use easy-peasy in a specific manner - and we don't have tests for persistance, RN, etc. - so there might be some outliers, but I'm hoping internal tests is covering this.

I have not looked into the details of the immer upgrade. If there was an issue in regards to the upgrade, I'm hoping that the tests (local and/or external) would uncover it - but I get all greens here, so I'm just assuming everything is good.

Have you tested these changes against your codebase @no-stack-dub-sack ?

no-stack-dub-sack:
I haven't but I don't think they'd gain us anything that your tests don't cover given how extensive your tests are and that my tests also won't cover the areas you cited. Let me take a quick look at the Immer change log but I think its probably good to go otherwise.

no-stack-dub-sack:
@jmyrland we may have some issues here: https://github.com/immerjs/immer/releases/tag/v10.0.0.

Specifically with the [proxy polyfill / support for older browsers](https://easy-peasy.dev/docs/introduction/browser-support.html#polyfilling-proxy). Immer has removed support for this entirely and recommends that projects that use `enableES5()` don't migrate, which we use in the polyfill. I think we either have to remove support for older browsers as well or don't migrate to v10.

jmyrland:
> @jmyrland we may have some issues here: https://github.com/immerjs/immer/releases/tag/v10.0.0.
> 
> Specifically with the [proxy polyfill / support for older browsers](https://easy-peasy.dev/docs/introduction/browser-support.html#polyfilling-proxy). Immer has removed support for this entirely and recommends that projects that use `enableES5()` don't migrate, which we [use for the polyfill](https://github.com/ctrlplusb/easy-peasy/blob/master/proxy-polyfill.js). I think we either have to remove support for older browsers as well or don't migrate to v10.

Ah I see! Good catch @no-stack-dub-sack. I'll downgrade it to the latest patch 👍

no-stack-dub-sack (approved):
LGTM!

PR #864, opened 2023-09-15
bugfix: Allow async effects for `effectOn`

jmyrland:
The docs states that [async effects should be allowed in an `effectOn`](https://easy-peasy.dev/docs/api/effect-on.html#utilising-a-dispose-function).

This works in JS, but not in TS - due to the type definition.

Updates the type defs to allow for async effects.


no-stack-dub-sack (approved):
LGTM!

PR #874, opened 2023-10-12
fix: computed props w/ dependencies on store state

### LINKED ISSUE #873: Computed props which depend on computed props that consume store state don't update as expected
See this CodeSandbox:

[Edit Dependent Computed Prop Issue Repro](https://codesandbox.io/s/dependent-computed-prop-issue-repro-rcvsgj?fontsize=14&hidenavigation=1&theme=dark)

The `readonly` computed property depends on `readonlyReason`. `readonlyReason` consumes state external to `bar` from the `foo` state slice.

When "Submit" is clicked, `readonly` should become `true` based on how the computed property is written. However it remains false. The current workaround is to define an explicit state resolver to isolate `readonlyReason`. When `readonlyReason` is accessed directly by the state resolver, `readonly` returns the expected value. See `store.ts` and try toggling the computed props to see this in action.

I believe what's happening here is that [this check](https://github.com/ctrlplusb/easy-peasy/blob/11903f863d43f4b3829d88faa3c237b316c380f5/src/computed-properties.js#L33) prevents the computed prop from re-computing any time the computed prop's inputs are the same _OR_ if store state is equal to the previous store state. In this case, the inputs are equal (with no explicit state resolvers, the default state resolvers of `[(state) => state]` are used, and no local state has changed), but store state and previous store state are different, and since the prop in question depends on a computed property that depends only on external state, its not updating correctly.

I have a possible fix in mind and will test with the help of @jmyrland before opening a PR. If for some reason this cannot be fixed, we should update the documentation to cite this as a known issue.

no-stack-dub-sack:
- Changes `storeState === prevStoreState || areInputsEqual(...)` to `storeState === prevStoreState && areInputsEqual(...)` when considering whether or not to return the cached value of a computed prop. Only if both conditions are true should we blindly return the cached value. If either condition is false, we should re-calculate and find out if the value has changed. If it has not, return the cached value still (this may happen if inputs are equal and store state has changed that the computed prop does not depend on). Otherwise update cached value and trigger re-render.
- I believe this is related to the changes introduced in #764. This PR optimized computed props to reduce/remove the number of unwanted re-renders, however the optimization was maybe a bit too aggressive and missed a few edge cases where it doesn't trigger a re-render when it should.
- The new test fails prior to this code change and illustrates the issue that prompted the filing of #873.
- This PR also removes an un-used argument from the `createComputedPropertyBinder` and changes the `runOnce` variable name to `hasRunOnce` (just for a little more clarity - this always confused be a bit at first).
 
fixes #873

### DIFF
--- a/src/computed-properties.js
+++ b/src/computed-properties.js
@@ -1,8 +1,8 @@
 import equal from 'fast-deep-equal/es6';
 import { areInputsEqual } from './lib';
 
-export function createComputedPropertyBinder(parentPath, key, def, _r) {
-  let runOnce = false;
+export function createComputedPropertyBinder(key, def, _r) {
+  let hasRunOnce = false;
   let prevInputs = [];
   let prevValue;
   let prevStoreState;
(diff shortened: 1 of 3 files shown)


jmyrland (approved):
LGTM 👌

PR #876, opened 2023-10-13
chore: minor, non-functional changes

no-stack-dub-sack:
- refactor: remove dead code
- refactor: move `handleEventDispatchErrors` to lib
- chore: fix typo
- chore: add npmrc file to configure registry

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+registry = "https://registry.npmjs.org"
--- a/src/extract-data-from-model.js
+++ b/src/extract-data-from-model.js
@@ -26,7 +26,6 @@ export default function extractDataFromModel(
   const _aCD = {};
   const _aC = {};
   const _aRD = {};
-  const actionThunks = {};
   const _cP = [];
   const _cR = [];
   const _e = [];
@@ -125,9 +124,6 @@ export default function extractDataFromModel(
           // Create the function that will handle, i.e. be executed, when
           // the thunk action is created/dispatched
           def.thunkHandler = createThunkHandler(def, _r, injections, _aC);
-
-          // Register the thunk handler
-          set(path, actionThunks, def.thunkHandler);
 
           // Create the "action creator" function
           def.actionCreator = createThunkActionsCreator(def, _r);
--- a/src/lib.js
+++ b/src/lib.js
@@ -212,3 +212,25 @@ export function useMemoOne(
 
   return cache.result;
 }
+
+const logEventListenerError = (type, err) => {
+  // eslint-disable-next-line no-console
+  console.log(`Error in ${type}`);
+  // eslint-disable-next-line no-console
+  console.log(err);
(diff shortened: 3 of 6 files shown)


src/extract-data-from-model.js
```diff
@@ -26,7 +26,6 @@ export default function extractDataFromModel(
   const _aCD = {};
   const _aC = {};
   const _aRD = {};
-  const actionThunks = {};
```
no-stack-dub-sack, line 29: As far as I can tell, this was never being used. All tests still pass.

src/lib.js
```diff
@@ -212,3 +212,25 @@ export function useMemoOne(
...
+      const result = dispatcher(...args);
+      if (isPromise(result)) {
+        result.catch((err) => {
+          logEventListenerError(type, err);
+        });
+      }
+    } catch (err) {
+      logEventListenerError(type, err);
+    }
+  };
```
no-stack-dub-sack, line 236: Make the code a bit more DRY by replacing repeated code with shared util.

src/persistence.js
```diff
@@ -7,7 +7,7 @@ const noopStorage = {
   removeItem: () => undefined,
 };
 
-const getBrowerStorage = (storageName) => {
+const getBrowserStorage = (storageName) => {
```
no-stack-dub-sack, line 10: Whoops!

.npmrc
no-stack-dub-sack, line 1: For devs that use private NPM registries. Avoid conflicts with commands like `npm publish`

src/extract-data-from-model.js
```diff
@@ -126,9 +125,6 @@ export default function extractDataFromModel(
           // the thunk action is created/dispatched
           def.thunkHandler = createThunkHandler(def, _r, injections, _aC);
 
-          // Register the thunk handler
-          set(path, actionThunks, def.thunkHandler);
-
```
jmyrland, line 131: Maybe we don't have tests covering this? IDK what `actionThunks` are

jmyrland: I've tested this branch against our codebase, and we get 0 failures - so I guess this is fine.

jmyrland (approved):
Great stuff 👌

src/extract-data-from-model.js
no-stack-dub-sack: Yeah, this is related to this change: https://github.com/ctrlplusb/easy-peasy/pull/876#discussion_r1358580832. The only place `actionThunks` was ever being used was here, the `thunkHandler` was being set on the object but then the object never got passed along to anything else. And if all the tests still pass, I'm thinking it's fine. Ctrl + Shift + F also turned up no other references to `actionThunks`. Obviously this isn't definitive because it could be passed to a func that calls it something else, but if deleting this line doesn't break anything else I think we're good.

PR #946, opened 2025-03-21
Fix typo in thunk.md

ws333:


jmyrland (approved):
Thanks for the contribution @ws333 ! Sadly the docs CI is broken, so it won't get published (unless @ctrlplusb fixes it).

I will however merge your contribution, so it will get updated in the future when the CI is fixed.

